PR #12, opened 2023-03-13
TransferBench Package

PedramAlizadeh:


CMakeLists.txt
```diff
@@ -13,3 +14,35 @@ set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ..)
...
+include(ROCMCreatePackage)
+set(ROCMCHECKS_WARN_TOOLCHAIN_VAR OFF)
+
+set(PACKAGE_NAME TB)
+set(LIBRARY_NAME TransferBench)
+
+rocm_package_setup_component(${PACKAGE_NAME}
+    LIBRARY_NAME ${LIBRARY_NAME}
+    PACKAGE_NAME ${PACKAGE_NAME}
+)
```
lawruble13: I don't think that this is necessary, the default packages should be sufficient.

CMakeLists.txt
```diff
@@ -13,3 +14,35 @@ set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ..)
...
+# FILES_MATCHING
+# PATTERN "*.hpp"
+# )
+
+rocm_install(DIRECTORY src/include
+DESTINATION include/${PACKAGE_NAME}
+COMPONENT ${LIBRARY_NAME}
+FILES_MATCHING
+PATTERN "*.hpp"
+)
```
lawruble13: As Transferbench only has the single executable, installing the header files shouldn't be necessary. However, you do need to ensure that the executable itself is installed.

CMakeLists.txt
```diff
@@ -13,3 +14,35 @@ set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ..)
...
+
+rocm_create_package(
+    NAME ${LIBRARY_NAME}
+    DESCRIPTION "TransferBench package"
+    MAINTAINER "RCCL Team"
```
lawruble13: An email should be included in angle brackets after the maintainer name.

lawruble13 (commented):
Looking pretty good, thanks Pedram! As TransferBench is fairly simple you've actually overdone it a bit here, see my comments for further information.

CMakeLists.txt
PedramAlizadeh: Hi Lauren,
Thank you very much for the feedbacks!
I applied the modifications. Does it look correct to you?
Also, for creating the packages after "cmake .. && make", I do "make package"!
Is it the correct way to do so?
Then, I get the following error for RPM:
CPackRPM:Debug: You may consult rpmbuild logs in: 
CPackRPM:Debug:    -TransferBench-dev/build/_CPack_Packages/Linux/RPM/rpmbuildtransferbench-devel.err
CPackRPM:Debug: *** error: line 9: Illegal sequence ".." in: Version:        ..

CMake Error at /snap/cmake/1270/share/cmake-3.26/Modules/Internal/CPack/CPackRPM.cmake:1897 (message):
  RPM package was not generated!
  TransferBench-dev/build/_CPack_Packages/Linux/RPM
Call Stack (most recent call first):
  /snap/cmake/1270/share/cmake-3.26/Modules/Internal/CPack/CPackRPM.cmake:1968 (cpack_rpm_generate_package)

CPack Error: Error while execution CPackRPM.cmake

PedramAlizadeh: I think the problem was VERSION in the project command!
It is solved now and seems to be working!

lawruble13 (approved):
LGTM!

PR #48, opened 2023-08-09
Adding standard deviation calculations

gilbertlee-amd:


src/TransferBench.cpp
```diff
@@ -385,15 +385,29 @@ void ExecuteTransfers(EnvVars const& ev,
...
+    double avg = 0;
+    for (auto timing : transfer->perIterationTiming)
+      avg += timing;
+    avg /= transfer->perIterationTiming.size();
+
+    double sumSquares = 0;
+    for (auto timing : transfer->perIterationTiming)
+      sumSquares += (timing - avg) * (timing - avg);
+    transfer->stdDev = sqrt(sumSquares / transfer->perIterationTiming.size());
+    avgStdDev += transfer->stdDev;
```
astatide, line 405: We ended up discussing it in chat, so I'm just putting it here for posterity, that it's probably better to report the standard deviations independently due to these being forward/reverse processes (and also that you can potentially get into nasty territory when averaging statistical quantities and that you wanna be extra careful that all the math works out and anyway, it's just easier to not).

wenkaidu (approved):

astatide (approved):

PedramAlizadeh (approved):

PR #103, opened 2024-04-29
Transfer benchdocs reorg

Rmalavally:


docs/index.rst
randyh62, line 1: I think I would just title this "TransferBench documentation" or something along those lines.

randyh62, line 1: I would title this "Using TransferBench" as opposed to "Use TransferBench".

randyh62 (approved):

docs/index.rst
Rmalavally: The header is How to, so 'Using' will be grammatically incorrect. I think we have 'Use' in our template as well. 

https://rocm.docs.amd.com/projects/rocDecode/en/latest/index.html

docs/index.rst
```diff
@@ -1,11 +1,36 @@
-*******************************************
-Welcome to TransferBench's documentation!
-*******************************************
+****************************
+TransferBench documentation
+****************************
 TransferBench is a simple utility capable of benchmarking simultaneous transfers between user-specified devices (CPUs/GPUs).
```
lpaoletti: Perhaps a little tidying up in the intro:
TransferBench is a utility capable of benchmarking simultaneous transfers between user-specified devices (CPUs/GPUs). A transfer is a single operation where an executor reads and adds values from source (SRC) memory locations, then writes the sum to destination (DST) memory locations. This simplifies to a simple copy operation when dealing with a single SRC/DST.

docs/index.rst
```diff
@@ -1,11 +1,36 @@
...
 A Transfer is defined as a single operation where an executor reads and adds together values from source (SRC) memory locations, then writes the sum to destination (DST) memory locations. This simplifies to a simple copy operation when dealing with single SRC/DST.
 
-The user has control over the SRC and DST memory locations by indicating memory type followed by the device index. TransferBench supports coarse-grained pinned host memory, unpinned host memory, fine-grained host memory, coarse-grained global device memory, fine-grained global device memory, and null memory (for an empty transfer). In addition, the user can determine the size of the transfer (number of bytes to copy) for their tests.
 
-The executor of the transfer can also be specified by the user. The options are CPU, kernel-based GPU, and SDMA-based GPU (DMA) executors. TransferBench also provides the option to choose the number of sub-executors. In case of a CPU executor this argument specifies the number of CPU threads, while for a GPU executor it defines the number of compute units (CU). If DMA is specified as the executor, the sub-executor argument determines the number of streams to be used.
+.. grid:: 2
```
lpaoletti, line 8: Should we have a link to the Git repository after the intro?

docs/install/install.rst
```diff
@@ -0,0 +1,49 @@
+.. meta::
+  :description: TransferBench documentation 
+  :keywords: TransferBench, API, ROCm, HIP
+
+---------------------------
+TransferBench Installation
+---------------------------
+
```
lpaoletti, line 8: Shouldn't there be an intro sentence before these two bullets?
Perhaps system requirements?

docs/how to/use-transferbench.rst
```diff
@@ -0,0 +1,22 @@
...
+* fine-grained host memory
+* coarse-grained global device memory
+* fine-grained global device memory
+* null memory (for an empty transfer).
+
+In addition, users can determine the size of the transfer (number of bytes to copy) for their tests.
+
+Users can also specify executors of the transfer. The options are CPU, kernel-based GPU, and SDMA-based GPU (DMA) executors. TransferBench also provides the option to choose the number of sub-executors. In case of a CPU executor this argument specifies the number of CPU threads, while for a GPU executor it defines the number of compute units (CU). If DMA is specified as the executor, the sub-executor argument determines the number of streams to be used.
+
+Refer to :ref:`Examples` for a specific example.
```
lpaoletti: We only have one example. This bounces the user to a page with a single link. 
Let's get rid of the page and have a new link. 
Something like:
See ConfigFile Format for an example of using TransferBench.

lpaoletti (commented):
The topic is called:
ConfigFile Format

However, the TOC is named:
Configfile format

They should align. 

============================
Taking a quick look at the Intro paragraph:
A Transfer is defined as a single operation where an Executor reads and adds together values from Source (SRC) memory locations, then writes the sum to destination (DST) memory locations. This simplifies to a simple copy operation when dealing with single SRC/DST.:

-paragraph ends with a period/colon
-Transfer uses an initial cap
-does this paragraph actually introduce what the topic is about? is this a tutorial?

docs/index.rst
Rmalavally: Fixed

docs/index.rst
Rmalavally: Fixed.

docs/install/install.rst
Rmalavally: We need the content from the Component team.

docs/how to/use-transferbench.rst
Rmalavally: Done.

Rmalavally:
Fixed.

docs/how to/use-transferbench.rst
Rmalavally: Thanks, Leo. I added the example to the Use TransferBench file for contextual reference.

gilbertlee-amd (approved):

PR #122, opened 2024-08-28
V1.52 candidate

gilbertlee-amd:
### Added
- Added USE_HSA_DMA env var to switch to using hsa_amd_memory_async_copy instead of hipMemcpyAsync for DMA execution
- Added ability to set USE_GPU_DMA env var for a2a benchmark


edgargabriel (approved):
Based on reading the code, it looks good to me.

gilbertlee-amd:
Adding a warning if large BAR is disabled
Skipping numa hop counting if HSA returns that there are 0 hops between GPU and CPU.

PR #130, opened 2024-10-02
RDMA executor support

mustafabar:
### RDMA Executor

1. Added support for RDMA-based data transfers in TransferBench. This executor enables data transfers using RDMA-capable NICs, bypassing host involvement for GPU/CPU transfers.
2. Memory registration function to handle memory registration required for RDMA transfers. Memory is preregistered so that the cost of registration is not included in the transfer
3. Added initialization (`InitDeviceAndQPs`) and teardown (`TearDown`) processes for RDMA Executors.
4. Added topology display support and a mode for automatically mapping GPUs to closest NICs for GPU course and fine-grained mem types 
5. Sub-executor support was added to express the number of QPs used per transfer. Data is split over the QPs
6. Automatic GID Index detection for RoCE traffic (Code extracted from recent NCCL support)
7. PCIe Tree is built and printed for NIC <->GPU mapping
8. Executor Rx indexes NIC x following ibverbs order, whereas executor Nx uses the nearest NIC to GPU index x
9. When topology detection is not accurate, especially in systems with front NICs and multiport NICs, the `NEAREST_NIC` variable can override the automatically detected topology.

### IBVerbs library Wrapper:

1. An IBVerbs wrapper to handle low-level RDMA operations
2. Logic added to detect and initialize RDMA NICs using IBVerbs (`InitDeviceAndQPs`) and CPU/GPU data transfers (`TransferData`).

### Fallback when IBVerbs is not available

1. Compile-time fallback added to the Makefile to switch to GPU-based DMA executors if RDMA support is unavailable on the dynamic library cache. Method stubs are included to avoid breaking compatibility. 
2. Updated the Makefile to conditionally compile RDMA executor support based on the availability of ibverbs on the system. If ibverbs is not found, RDMA executors are disabled with `LIB_IBVERBS_UNAVAILABLE`.

### Misc:

1. Added numNicDevices to the EnvVars structure, which stores the number of detected RDMA NICs to ensure that the correct number of NICs is used.
2. Stopped tracking binaries and ignored changes to binary files.
3. Added environment variable handling to control RDMA executor behavior, including the number of NICs to use (`NUM_RDMA_DEVICES`) and executor configurations via EnvVars.
4. Added input validation for RDMA NIC device counts. An error is raised if an invalid NIC index is specified, preventing out-of-bound access or using inactive NICs


mustafabar:
PR is only tested on a Mellanox IB server. Broadcom RoCE seems to have initialization issues. I am currently looking into it.

mustafabar:
It can be tested now. Please report any issues you encounter.

PR #147, opened 2024-12-02
RDMA Executor support

mustafabar:
### Added

1. Support for RDMA-based data transfers in TransferBench. This executor enables data transfers using RDMA-capable NICs, bypassing host involvement for GPU/CPU transfers
2. Two new executor symbols are added: `I` and `N`. `Nj` is the nearest NIC to `GPUj`. `Ij` is NIC with index `j`.
3. Memory is preregistered so that the cost of registration is not included in the transfer
4. Added topology display support using PCIe tree mapping algorithm
5. Multi-port NIC mapping support so that GPUs that share a PCIe bridge with a GPU are evenly distributed on the ports
6. QPs are expressed as the number of sub-executors per transfer. Data is split over the QPs
7. Automatic GID Index detection for RoCE traffic (Code extracted from recent NCCL support)
8. When topology detection is not accurate, especially in systems with front NICs and multiport NICs, the `CLOSEST_NIC` variable can override the automatically detected topology


src/client/Client.cpp
```diff
@@ -232,13 +233,25 @@ void PrintResults(EnvVars const& ev, int const testNum,
...
         sprintf(exeSubIndexStr, ".%d", t.exeSubIndex);
 
-      printf("     Transfer %02d  %c %7.3f GB/s %c %8.3f ms %c %12lu bytes %c %s -> %s%02d%s:%03d -> %s\n",
-             idx,                    sep,
-             r.avgBandwidthGbPerSec, sep,
-             r.avgDurationMsec,      sep,
-             r.numBytes,             sep,
-             MemDevicesToStr(t.srcs).c_str(), ExeTypeName[exeType], exeIndex,
-             exeSubIndexStr, t.numSubExecs, MemDevicesToStr(t.dsts).c_str());
+      if(exeType == EXE_IBV || exeType == EXE_IBV_NEAREST) {
```
gilbertlee-amd: Let's wrap up the creation of the executor string into a function and keep this common instead.

src/client/EnvVars.hpp
```diff
@@ -100,6 +100,13 @@ class EnvVars
   int outputToCsv;                   // Output in CSV format
   int samplingFactor;                // Affects how many different values of N are generated (when N set to 0)
 
+  // Rdma options
+  int ibGidIndex;                    // GID Index for RoCE NICs
+  int roceVersion;                   // RoCE version number
+  int ipAddressFamily;               // IP Address Famliy
+  uint8_t ibPort;                    // NIC port number to be used
+  std::string closestNicStr;            // Holds the user-specified list of closest NICs
```
gilbertlee-amd: spacing

src/client/EnvVars.hpp
```diff
@@ -299,6 +313,11 @@ class EnvVars
     printf(" USE_SINGLE_STREAM - Use a single stream per GPU GFX executor instead of stream per Transfer\n");
     printf(" VALIDATE_DIRECT   - Validate GPU destination memory directly instead of staging GPU memory on host\n");
     printf(" VALIDATE_SOURCE   - Validate GPU src memory immediately after preparation\n");
+    printf(" IB_GID_INDEX      - Required for RoCE NICs (default=3)\n");
```
gilbertlee-amd: default appears to be -1 though?

src/client/EnvVars.hpp
```diff
@@ -299,6 +313,11 @@ class EnvVars
...
+    printf(" IB_PORT_NUMBER    - RDMA port count for RDMA NIC (default=1)\n");
+    printf(" IP_ADDRESS_FAMILY - IP address family (4=v4, 6=v6, default=v4)\n");
+    printf(" ROCE_VERSION      - RoCE version (default=2)\n");
+    printf(" CLOSEST_NIC       - Comma seperated list of per-GPU closest NIC (default=auto)\n");
```
gilbertlee-amd: separated

src/client/EnvVars.hpp
```diff
@@ -479,6 +508,29 @@ class EnvVars
...
+    cfg.rdma.ipAddressFamily       = ipAddressFamily;
+    cfg.rdma.roceVersion           = roceVersion;
+    std::vector<int> closestNics;
+    if(closestNicStr != "") {      
+      std::stringstream ss(closestNicStr);
+      std::string item;
+      while (std::getline(ss, item, ',')) {
+        try {
+          int nic = std::stoi(item);
+          if (nic < 0) {
```
gilbertlee-amd: Do you also need to check that this is within the range of NICs?

CMakeLists.txt
```diff
@@ -56,6 +56,14 @@ set( CMAKE_CXX_FLAGS "${flags_str}${CMAKE_CXX_FLAGS}")
 
 set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -L${ROCM_PATH}/lib")
 include_directories(${ROCM_PATH}/include)
+find_library(IBVERBS_LIBRARY ibverbs)
+if (IBVERBS_LIBRARY)
+    message(STATUS "Found ibverbs: ${IBVERBS_LIBRARY}")
+    link_libraries(ibverbs)    
+else()
+    add_definitions(-DNO_IBV_EXEC)
```
gilbertlee-amd: Can we change this to -DIBV_EXEC_ENABLED?  
A lot of the following code uses #ifndef NO_IBV_EXEC and the double negative seems harder to immediately understand versus #ifdef IBV_EXEC_ENABLED

src/client/Topology.hpp
```diff
@@ -38,21 +43,107 @@ static int RemappedCpuIndex(int origIdx)
...
+              if (ibv_query_port(ctx, port, &portAttr) == 0 && portAttr.state == IBV_PORT_ACTIVE) {
+                devPortsActive[i] = true;
+                break;              
+              }
+            }
+          }
+
+          std::string devicePath(deviceList[i]->dev_path);
+          if (std::filesystem::exists(devicePath))
+          {
```
gilbertlee-amd: bracket

src/client/Topology.hpp
```diff
@@ -38,21 +43,107 @@ static int RemappedCpuIndex(int origIdx)
...
+              devBusIds[i] = nicBusId;              
+            }
+          }
+          ibv_close_device(ctx);
+        }
+      }
+      ibv_free_device_list(deviceList);
+    }
+    for (int i = 0; i < devBusIds.size(); ++i)
+    {
```
gilbertlee-amd: bracket

src/client/Topology.hpp
```diff
@@ -38,21 +43,107 @@ static int RemappedCpuIndex(int origIdx)
...
+               closestGpusStr.c_str(), 
+               devBusIds[i].c_str());
+      }
+      else {
+        printf("%-12d | %-11s | %-11s | %-13s | %-11s\n", 
+               i, 
+               nicDevice.c_str(), 
+               portActive ? "Yes" : "No", 
+               closestGpusStr.c_str(), 
+               devBusIds[i].c_str());
```
gilbertlee-amd: Let's use separator as either ',' or '|' and combine these two

src/client/Topology.hpp
```diff
@@ -38,21 +43,107 @@ static int RemappedCpuIndex(int origIdx)
...
+    for (int i = 0; i < devBusIds.size(); ++i)
+    {
+      std::string nicDevice = devNames[i];
+      bool portActive = devPortsActive[i];
+
+      auto closestGpus = GetClosestGpusToNic(i);
+      std::string closestGpusStr;
+      for (size_t j = 0; j < closestGpus.size(); ++j) {
+        closestGpusStr += std::to_string(closestGpus[j]);
+        if (j < closestGpus.size() - 1) {
```
gilbertlee-amd: Recommend putting this before 103 instead.
if (j) closestGpuStr += ",";

src/header/TransferBench.hpp
```diff
@@ -63,10 +76,12 @@ namespace TransferBench
     EXE_GPU_GFX      = 1,                       ///<  GPU kernel-based executor (subExecutor = threadblock/CU)
     EXE_GPU_DMA      = 2,                       ///<  GPU SDMA executor         (subExecutor = not supported)
     EXE_IBV          = 3,                       ///<  IBVerbs executor          (subExecutor = queue pair)
+    EXE_IBV_NEAREST  = 4                        ///<  IBVerbs nearest executor  (subExecutor = queue pair)
```
gilbertlee-amd: Let's change this to EXE_NIC and EXE_NIC_NEAREST so that we don't need to change the API when we start supporting different RDMA providers.

src/client/Topology.hpp
```diff
@@ -38,21 +43,107 @@ static int RemappedCpuIndex(int origIdx)
...
+        }
+      }
+      ibv_free_device_list(deviceList);
+    }
+    for (int i = 0; i < devBusIds.size(); ++i)
+    {
+      std::string nicDevice = devNames[i];
+      bool portActive = devPortsActive[i];
+
+      auto closestGpus = GetClosestGpusToNic(i);
```
gilbertlee-amd: We want client to be an example of how TransferBench library can be used
Let's add TransferBench::GetClosestGpusToNic(i) here to make it clear that this is a call provided by TransferBench

src/client/EnvVars.hpp
```diff
@@ -479,6 +508,29 @@ class EnvVars
...
+            printf("[ERROR]: NIC index cannot be negative\n");
+            exit(1);
+          }
+          closestNics.push_back(nic);
+        } catch (const std::invalid_argument& e) {
+          printf("[ERROR] Invalid NIC index: %s\n", item.c_str());
+          exit(1);
+        }
+      }
+      TransferBench::SetClosestNics(closestNics);
```
gilbertlee-amd: Can this be reworked and passed into ConfigOptions instead?  
This breaks with the current conventions.
I'd rather that we pass configOptions into TransferBench::GetClosestGpusToNic if necessary.

src/header/TransferBench.hpp
```diff
@@ -457,6 +508,26 @@ namespace TransferBench
...
+// Helper macro for calling RDMA functions and catching nullptr errors
+#define IBV_PTR_CALL(__ptr__, __func__, ...)                            \
+  do {                                                                  \
+    __ptr__ = __func__(__VA_ARGS__);                                    \
+    if (__ptr__ == nullptr) {                                           \
+      return {ERR_FATAL, "Encountered IbVerbs nullptr error at line (%d) " \
+              "and function (%s)", __LINE__, #__func__};                \
+    }                                                                   \
+  } while (0)
+
```
gilbertlee-amd: As this is expected to be integrated with other tools, we may need to be careful about displaying line information.

Let's only enable this for debug builds via a define

src/header/TransferBench.hpp
```diff
@@ -1352,7 +1448,17 @@ namespace {
...
+static ErrResult InitRdmaTransferResources(Transfer          const& transfer,
+                                           TransferResources &      resources,
+                                           RdmaOptions       const& rdmaOptions);
+static ErrResult RegisterRdmaMemoryTransfer(Transfer          const& transfer,
+                                            TransferResources &      resources,
+                                            int               & transferId);
+static ErrResult TransferRdma(TransferResources & resources,
+                              int               const& transferIdx);
+static ErrResult TeardownRdma(TransferResources & resources);
+#endif
```
gilbertlee-amd: We shouldn't need to forward declare.  You can create a section for RDMA helper functions above (see similar HSA-related functions) section.

src/header/TransferBench.hpp
```diff
@@ -1487,7 +1593,26 @@ namespace {
...
     }
-
+    if (IsRdmaExeType(exeDevice.exeType)) {
+#ifndef NO_IBV_EXEC
+    for (auto& resources : exeInfo.resources) {
+      Transfer const& t = transfers[resources.transferIdx];
+      ERR_CHECK(InitRdmaTransferResources(t,
+                                          resources,
+                                          cfg.rdma
+                                          ));
```
gilbertlee-amd: args here can fit on single line

src/header/TransferBench.hpp
```diff
@@ -1155,6 +1238,19 @@ namespace {
     hsa_amd_sdma_engine_id_t   sdmaEngineId;      ///< DMA engine ID
 #endif
 
+#ifndef NO_IBV_EXEC
+    vector<NicResources*> rdmaResourceMapper;    ///< Store resoruce sensitive RDMA fields
```
gilbertlee-amd: resoruce

src/header/TransferBench.hpp
```diff
@@ -1155,6 +1238,19 @@ namespace {
     hsa_amd_sdma_engine_id_t   sdmaEngineId;      ///< DMA engine ID
 #endif
 
+#ifndef NO_IBV_EXEC
```
gilbertlee-amd: For consistency, you can add the comment that this section is
// For IBV executor

src/header/TransferBench.hpp
```diff
@@ -2088,6 +2272,1114 @@ namespace {
...
+    IBV_PTR_CALL(rdma->completionQueue, ibv_create_cq, rdma->context, 100, NULL, NULL, 0);
+    IBV_CALL(ibv_query_port, rdma->context, port, &rdma->portAttr);
+    if (rdma->portAttr.state != IBV_PORT_ACTIVE) {      
+      return {ERR_FATAL, "Selected RDMA device %d is down", deviceID};
+    }
+  }
+  return ERR_NONE;
+}
+
+static ErrResult InitRdmaTransferResources(Transfer          const& transfer,
```
gilbertlee-amd: By convention, we place input arguments before in/out, output arguments.
I'd order this with rdmaOptions first, followed by Transfer, followed by resources.

src/header/TransferBench.hpp
```diff
@@ -2088,6 +2272,1114 @@ namespace {
...
+    if (err.errType != ERR_NONE) {
+      return err;
+    }
+  }
+  return ERR_NONE;
+}
+
+static ErrResult InitDeviceList() 
+{
+  if (deviceList == NULL)
```
gilbertlee-amd: bracket

src/header/TransferBench.hpp
```diff
@@ -2088,6 +2272,1114 @@ namespace {
...
+  } while(0);
+
+const  uint64_t WR_ID = 1789;
+static ibv_device** deviceList = nullptr;
+static int RdmaNicCount        = -1;
+static std::vector<std::string> IbDeviceBusIds;
+static std::vector<std::set<int>> NicToGpuMapper;
+static std::vector<int> GpuToNicMapper;
+static std::vector<std::string> DeviceNames;
+static int GpuCount;
```
gilbertlee-amd: GpuCount should be replaced by just calls to TransferBench::GetNumExecutors(EXE_GPU_GFX);

src/header/TransferBench.hpp
```diff
@@ -2088,6 +2272,1114 @@ namespace {
...
+static bool DeviceMappingInit  = false;
+static bool PcieTreeInit = false;
+static bool MultiportFlag = false;
```
gilbertlee-amd: naming convention is UpperCamelCase for functions, lowerCamelCase for variables.
We may need to be careful here as these variables are still going to be accessible by anyone including the header file.  Perhaps add underscore prefix, or take a second look to see if these are all necessary to be "global".

I think the better approach should have a accessor function that captures this with static variables within the function instead.

e.g.:
ErrResult GetIbvDevice(int devIndex, ibv_device*& ibvDevice) {
  static ibv_device** deviceList = nullptr;
  if (deviceList == nullptr) {
   // initialize
   }
   // Error check for valid devIndex
   ...
}

src/header/TransferBench.hpp
```diff
@@ -2088,6 +2272,1114 @@ namespace {
...
+};
+
+static PCIe_tree pcie_root;
+
+static void InsertPCIePathToTree(PCIe_tree* root, const std::string& pcieAddress, const std::string& description)
+{
+  std::filesystem::path devicePath = "/sys/bus/pci/devices/" + pcieAddress;
+  if (!std::filesystem::exists(devicePath))
+  {
+    printf("[ERROR] Device path %s does not exist\n", devicePath.c_str());
```
gilbertlee-amd: We can't do printfs.  Will need to properly propagate error up.

src/header/TransferBench.hpp
```diff
@@ -2088,6 +2272,1114 @@ namespace {
...
+  for (auto&& child : node->children)
+  {
+    int distance = GetLcaDepth(targetBusID, &child, depth + 1);
+    if (distance != -1)
+    {
+      return distance;
+    }
+  }
+  return -1;
+}
```
gilbertlee-amd: bracket formatting

src/header/TransferBench.hpp
```diff
@@ -2088,6 +2272,1114 @@ namespace {
...
+  char delimiter;
+
+  std::istringstream iss(pcieAddress);
+  iss >> std::hex >> domain >> delimiter >> bus >> delimiter >> device >> delimiter >> function;
+
+  if (iss.fail())
+  {
+    printf("Invalid PCIe address format: %s\n", pcieAddress.c_str());
+    return -1; // Invalid bus number
+  }
```
gilbertlee-amd: bracket

src/header/TransferBench.hpp
```diff
@@ -2088,6 +2272,1114 @@ namespace {
...
+
+static int GetNearestPcieDeviceInTree(PCIe_tree                const& root, 
+                                      std::string              const& busID, 
+                                      std::vector<std::string> const& targetBusIds)
+{
+  int max_depth = -1;
+  int index_of_closest = -1;
+  std::vector <int> matches;
+  for (const auto& targetBusID : targetBusIds)
+  {
```
gilbertlee-amd: bracket

src/header/TransferBench.hpp
```diff
@@ -2088,6 +2272,1114 @@ namespace {
...
+        max_depth = depth;
+        index_of_closest = &targetBusID - &targetBusIds[0];
+        matches.clear(); // found a new max depth
+        matches.push_back(index_of_closest);
+      }
+      else if(depth == max_depth && depth >= 0)
+      {
+        matches.push_back(&targetBusID - &targetBusIds[0]);
+      }
+    }
```
gilbertlee-amd: bracketing

gilbertlee-amd (changes requested):
Initial review pass.

CMakeLists.txt
mustafabar: I agree this is counterintuitive. My only reason for doing so is that I have noted that many folks, including me recently, use VS code + Intellisense. Since Intellisense doesn't see it as defined, huge portions of the code appear grayed out. Alternatively, we can track a .vscode/settings.json file where we define the environment.

corey-derochie-amd: I'd recommend writing the code without the negative and individuals can define things in their own environment.

mustafabar: another reason is usability: so I didn't want to have users include an extra CXXFLAG in the common use-cases, where NIC is already available. They will only need to add an extra flag to avoid using it, which is less common. In reality, all of this should be automated with autoconfig/CMake, but in the case of a header-only library, I feel the building/linking burden is pushed on the user.

corey-derochie-amd: If you wanted to satisfy both, you could use `IBV_EXEC_ENABLED` throughout the code, then at top of header use `#ifndef NO_IBV_EXEC #define IBV_EXEC_ENABLED #endif`. But I think either is probably fine.

src/client/Client.cpp
mustafabar: Please check latest

src/client/EnvVars.hpp
mustafabar: Fixed

src/client/EnvVars.hpp
mustafabar: Fixed

src/client/EnvVars.hpp
mustafabar: done

src/header/TransferBench.hpp
mustafabar: done

src/client/EnvVars.hpp
mustafabar: This is reworked now. Please check

src/header/TransferBench.hpp
mustafabar: Thanks for the suggested workaround for globals. I think this can be its own PR. I have temporarily used an underscore prefix and adjusted the naming convention

CMakeLists.txt
mustafabar: > If you wanted to satisfy both, you could use IBV_EXEC_ENABLED throughout the code, then at top of header use #ifndef NO_IBV_EXEC #define IBV_EXEC_ENABLED #endif. But I think either is probably fine.

I like this idea @corey-derochie-amd. I think when we cover Slingshot NICs, opting out with a disable macro will make more sense than disabling NIC exec by default

src/header/TransferBench.hpp
mustafabar: Resolved by merging the recent PR from @gilbertlee-amd's branch

gilbertlee-amd:
Duplicate of v1.59 candidate

PR #167, opened 2025-02-05
Add alltoall network preset

mustafabar:
### Added

- Added a2a_n preset which conducts alltoall GPU-to-GPU tranfers over nearest NIC executors


gilbertlee-amd (changes requested):
Looks good, however please increment client version by (as this doesn't affect library), and update CHANGELOG.

gilbertlee-amd (approved):

PR #172, opened 2025-02-21
Add priority ordering for GID index detection for RoCEv2 and rewrite logic

mustafabar:
### Fixed:
- Have a priority list of which GID entry to go for instead of hardcoding choices based on underdocumented user input (such as RoCE version and IP address family)
- Use link-local when it is the only choice (i.e. when routing information is not available beyond local link)

### Changed:
- Introduce maintainable and reusable code for main logic and IPv4-mapped addresses.
- Show the GID index in the topology table. It is helpful for debugging purposes.

### DIFF
--- a/src/header/TransferBench.hpp
+++ b/src/header/TransferBench.hpp
@@ -1801,9 +1801,9 @@ namespace {
     return true;
   }
 
-  static bool LinkLocalGid(union ibv_gid* gid)
+  static bool LinkLocalGid(union ibv_gid const& gid)
   {
-    const struct in6_addr *a = (struct in6_addr *)gid->raw;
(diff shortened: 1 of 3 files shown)


gilbertlee-amd (approved):

src/header/TransferBench.hpp
```diff
@@ -1891,60 +1834,95 @@ namespace {
...
-        gidCurrFam         = gidTestFam;
-        gidCurrIsValid     = true;
-        gidCurrRoceVersion = gidTestRoceVersion;
-      }
-    }
-
-    gidIndex = gidIndexCurr;
+                               struct ibv_context* context,
+                              const int& gidTblLen,
+                              int& gidIndex)
```
gilbertlee-amd: spacing here seems inconsistent

PR #186, opened 2025-06-13
Disable NIC executor by default for CMake builds

gilbertlee-amd:
Adding additional explicit opt-in for NIC executor for CMake builds


CMakeLists.txt
```diff
@@ -56,14 +56,19 @@ set( CMAKE_CXX_FLAGS "${flags_str} ${CMAKE_CXX_FLAGS}")
 
 set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -L${ROCM_PATH}/lib")
 include_directories(${ROCM_PATH}/include)
+
 find_library(IBVERBS_LIBRARY ibverbs)
 find_path(IBVERBS_INCLUDE_DIR infiniband/verbs.h)
 if (DEFINED ENV{DISABLE_NIC_EXEC})
   message(STATUS "Disabling NIC Executor support")
 elseif(IBVERBS_LIBRARY AND IBVERBS_INCLUDE_DIR)
   message(STATUS "Found ibverbs: ${IBVERBS_LIBRARY}. Building with NIC executor support. Can set DISABLE_NIC_EXEC=1 to disable")
```
mustafabar, line 65: Need to adjust the message to opt-in instead of opt-out. Suggested in another comment

CMakeLists.txt
```diff
@@ -56,14 +56,19 @@ set( CMAKE_CXX_FLAGS "${flags_str} ${CMAKE_CXX_FLAGS}")
...
-  add_definitions(-DNIC_EXEC_ENABLED)
-  link_libraries(ibverbs)
+  if (DEFINED ENV{ENABLE_NIC_EXEC})
+    add_definitions(-DNIC_EXEC_ENABLED)
+    link_libraries(ibverbs)
+  else()
+    message(STATUS "For CMake builds, NIC executor also requires explicit opt-in by setting NIC_EXEC_ENABLE=1")
```
mustafabar, line 70: Align with actual env
```suggestion
    message(STATUS "For CMake builds, NIC executor also requires explicit opt-in by setting ENABLE_NIC_EXEC=1")
```

mustafabar (changes requested):

CMakeLists.txt
mustafabar, line 65:
```suggestion
  message(STATUS "Found ibverbs: ${IBVERBS_LIBRARY}")
```

PR #195, opened 2025-08-14
[BUILD] Fix CMake logic to detect DISABLE_NIC_EXEC

nileshnegi:
## Motivation
- CMake builds should disable NIC executor if env. flag `DISABLE_NIC_EXEC` is set to 1.
- Additionally, CMake flag `ENABLE_NIC_EXEC` should be explicitly enabled to build with NIC executor support.

## Technical Details
- Modified CMake to not just check for the existence of `DISABLE_NIC_EXEC`, but also if it is enabled.
- Modified CMake to check if cmake flag `ENABLE_NIC_EXEC` is enabled.

## Test Plan
Verified with baremetal and docker runs but setting `export DISABLE_NIC_EXEC=1`

## Test Result
CMake builds disabled NIC executor

## Submission Checklist


thananon (approved):

thananon:
I also tested on Debian 10.

PR #198, opened 2025-09-03
v1.64.00

gilbertlee-amd:
### Added
- Added BLOCKSIZES to a2asweep preset to allow also sweeping over threadblock sizes
- Added FILL_COMPRESS to allow more control over input data pattern
  - FILL_COMPRESS takes in a comma-separated list of integer percentages (that must add up to 100)
    that sets the percentages of 64B lines to be filled by random/1B0/2B0/4B0/32B0 data patterns
    - Bins:
      - 0 - random
      - 1 - 1B0    upper 1 byte of each aligned 2 bytes is 0
      - 2 - 2B0    upper 2 bytes of each aligned 4 bytes is 0
      - 3 - 4B0    upper 4 bytes of each aligned 8 bytes is 0
      - 4 - 32B0   upper 32 bytes of each aligned 64-byte line are 0
  - FILL_PATTERN will be ignored if FILL_COMPRESS is specified
- Additional details about data patterns generated will be printed if the debug env var DUMP_LINES is
  set to a non-zero value, which also corresponds to how many 64 byte lines will be printed
### Modified
- Increased GFX_BLOCKSIZE limit from 512 to 1024 (still requires multiple of 64)

### Fixed
- Fixed bug when using BYTE_OFFSET


src/header/TransferBench.hpp
```diff
@@ -2162,8 +2176,104 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
...
-    if (bufferIdx >= 0) {
+    if (!cfg.data.fillCompress.empty()) {
+      // 0 -> Random
+      // 1 ->  1B0 - The upper  1 byte  of each aligned 2 bytes is 0
+      // 2 ->  2B0 - The upper  2 bytes of each aligned 4 bytes are 0
+      // 3 ->  4B0 - The upper  4 bytes of each aligned 8 bytes are 0
+      // 4 -> 32B0 - The upper 32 bytes of each 64-byte line are 0
+
+      // Fill buffer with random floats
+      std::default_random_engine gen;
```
mustafabar: std::default_random_engine is implementation-defined (not portable). If needed, std::mt19937 (or std::mt19937_64) can be used for stable cross-platform determinism.

src/header/TransferBench.hpp
```diff
@@ -2162,8 +2176,104 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
...
+      size_t leftover = numLines;
+      std::vector<size_t> lineCounts(5, 0);
+      std::set<std::pair<double, int>> remainder;
+
+      // Assign rounded down values first
+      std::vector<int> percentages = cfg.data.fillCompress;
+      while (percentages.size() < 5) percentages.push_back(0);
+      for (int i = 0; i < percentages.size(); i++){
+        lineCounts[i] = (size_t)(numLines * (percentages[i] / 100.0));
+        leftover -= lineCounts[i];
```
mustafabar, line 2205: Can we add an assertion to warn against silent off-by-one wrap around due to subtraction?

mustafabar (approved):

src/header/TransferBench.hpp
gilbertlee-amd: This shouldn't be necessary due to how casting rounds.

src/header/TransferBench.hpp
gilbertlee-amd: Thanks - changed.

PR #202, opened 2025-10-02
Check nvidia-smi output instead of just searching nvcc bin

mustafabar:
## Motivation

`make` thinks it is on NVIDIA hw when `nvcc` is found. Example: `nvcc` is found on an internal cluster with AMD hw because it happens to be on this path `/usr/local/cuda/bin/`

## Technical Details
Add an additional check:
If `nvidia-smi` is found and exited with 0 error code, the condition evaluates to true.
If `nvidia-smi` is not found, or found and exited with non-zero error code, the condition evaluates to false.

## Submission Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ GPU_TARGETS ?= native
 DEBUG ?= 0
 
 ifeq ($(filter clean,$(MAKECMDGOALS)),)
-  # Compile TransferBenchCuda if nvidia-smi returns successfully
+  # Compile TransferBenchCuda if nvidia-smi returns successfully and nvcc detected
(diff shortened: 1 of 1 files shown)


nileshnegi:
this seems like a decent approach... i had tried in the past to re-order the Makefile, to move the ROCm-related flags before the NVCC flags... which was not fruitful... would you say sticking to `nvidia-smi` is the best route?

nileshnegi (approved):

mustafabar:
> this seems like a decent approach... i had tried in the past to re-order the Makefile, to move the ROCm-related flags before the NVCC flags... which was not fruitful... would you say sticking to `nvidia-smi` is the best route?

Only problem with the `nvidia-smi` approach is when someone tries to cross compile without having the full CUDA stack. I don't have a better idea for now. As you said, reordering the conditions is a bit messy.

PR #212, opened 2025-11-12
v1.65 Adding warp-subexecutor support via GFX_SE_TYPE

gilbertlee-amd:
## Motivation
This PR adds a new execution method into TransferBench for GFX executors.
Each Transfer can now be assigned to a single warp instead of a single threadblock, which allows investigations on different approaches.  

## Technical Details
GFX_SE_TYPE has been introduced to allow switching between the two subexecutor modes.  
GFX_SE_TYPE=0 uses threadblocks as subexecutor (default), and GFX_SE_TYPE=1 uses warps as subexecutors.
This flag is passed all the way down to the kernel, which chooses which Transfer to work on either using warp index or threadblock index.

## Test Plan
All the previous validation is still maintained with warp subexecutors - Each Transfer has still validated at the end of all iterations.  AMG_LOG_LEVEL has been used to validate that the correct number of threadblocks are being launched.

## Test Result
A2A/P2P tests pass, with no performance regression.  Some performance improvements have also been noticed with warp subexecutors

## Submission Checklist


src/header/TransferBench.hpp
```diff
@@ -608,6 +610,38 @@ namespace {
...
+  int GetWarpSize(std::vector<ErrResult>* errors = nullptr) {
+    int warpSize = 0;
+    hipError_t err = hipDeviceGetAttribute(&warpSize, hipDeviceAttributeWarpSize, 0);
+    if (err == hipSuccess) {
+      return warpSize;
+    }
+
+    // Query failed, report error and fall back to compile-time default
+    if (errors) {
+      errors->push_back({ERR_WARN,
```
alex-breslow-amd, line 622: I think you're better off just exiting rather than falling back to a compile time default.  If the program fails in this way, it probably doesn't make sense to continue, especially since some AMD GPUs have a warp size of 32 instead of 64.

src/header/TransferBench.hpp
```diff
@@ -608,6 +610,38 @@ namespace {
...
+    return 64;
+#endif
+  }
+
+  // Calculate grid Y dimension based on SE_TYPE
+  int CalculateGridY(int seType, int blockSize, int numSubExecs) {
+    // Warp-level: each subexecutor is a warp, pack warps into threadblocks
+    if (seType == 1) {
+      int warpsPerBlock = blockSize / GetWarpSize();
+      return (numSubExecs + warpsPerBlock - 1) / warpsPerBlock;
```
alex-breslow-amd, line 638: Not sure if this is on the critical path.  I think not given it's CPU code, but if it is, you can get by with a right shift since you're dividing by a power of two (at least on 637).  I think there's similar tricks for 638.

src/header/TransferBench.hpp
```diff
@@ -1303,11 +1337,17 @@ namespace {
       {
         // Check total number of subexecutors requested
         int numGpuSubExec = GetNumSubExecutors(exeDevice);
+        // For warp-level dispatch, multiply by warps per threadblock
+        if (cfg.gfx.seType == 1) {
+          int warpsPerBlock = cfg.gfx.blockSize / GetWarpSize(&errors);
```
alex-breslow-amd, line 1342: Same comment here.  Consider memoizing GetWarpSize if it's on the critical path.  Again, I think not likely.

src/header/TransferBench.hpp
```diff
@@ -2977,12 +3017,29 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
   // Kernel for GFX execution
   template <typename PACKED_FLOAT, int BLOCKSIZE, int UNROLL, int TEMPORAL_MODE>
   __global__ void __launch_bounds__(BLOCKSIZE)
-    GpuReduceKernel(SubExecParam* params, int waveOrder, int numSubIterations)
+    GpuReduceKernel(SubExecParam* params, int seType, int waveOrder, int numSubIterations)
```
alex-breslow-amd, line 3020: Have you seen what the performance impact is from not making seType a template parameter?  I don't think it's likely much, probably about 1% to 2% if I were to guess, but I thought I'd mention it.

src/header/TransferBench.hpp
```diff
@@ -3106,8 +3171,15 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
     }
 
     // Wait for all threads to finish
-    __syncthreads();
-    if (threadIdx.x == 0) {
+    if (seType == 1) {
+      // For warp-level, sync within warp only
+      __syncwarp();
```
alex-breslow-amd, line 3176: I don't think you explicitly need this, but it is harmless.  It may prevent reordering by the compiler though, so that is beneficial.

alex-breslow-amd (approved):

src/header/TransferBench.hpp
gilbertlee-amd: It was originally templated.  There was no noticeable performance difference when switching over, but saved considerable amount of compilation time.

src/header/TransferBench.hpp
gilbertlee-amd: Likely still important because we still want to be able to run properly on NVIDIA hardware.

src/header/TransferBench.hpp
gilbertlee-amd: CPU code not even used during timing portion.

src/header/TransferBench.hpp
alex-breslow-amd: Aha, that makes sense.  I think it serving as a code motion barrier within a cuda thread is also important now that I think about it.  This is more idiomatic than our use of atomic signal fences in RCCL.

src/header/TransferBench.hpp
alex-breslow-amd: That makes sense then.

src/header/TransferBench.hpp
alex-breslow-amd: Should it be though in some cases?  I think this means that TransferBench will tend toward overly optimistic perf. scaling, but I guess that's okay if you're mainly interested on what happens on the GPUs.

PR #224, opened 2025-12-12
TransferBench v1.66 - Multi-Rank support

gilbertlee-amd:
## Motivation
This set of commits introduces multi-rank support to TransferBench

## Technical Details
Modifications made to support MPI or socket based backbone communication to facilitate multi-rank Transfers.
NIC executors can now be used to copy data between different ranks on different nodes.

## Submission Checklist


src/header/TransferBench.hpp
```diff
@@ -1979,19 +2666,19 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
     attr.min_rnr_timer      = 12;
     if (isRoCE) {
       attr.ah_attr.is_global                     = 1;
-      attr.ah_attr.grh.dgid.global.subnet_prefix = gid.global.subnet_prefix;
-      attr.ah_attr.grh.dgid.global.interface_id  = gid.global.interface_id;
+      attr.ah_attr.grh.dgid.global.subnet_prefix = connInfo.gid.global.subnet_prefix;
+      attr.ah_attr.grh.dgid.global.interface_id  = connInfo.gid.global.interface_id;
       attr.ah_attr.grh.flow_label                = 0;
-      attr.ah_attr.grh.sgid_index                = gidIndex;
+      attr.ah_attr.grh.sgid_index                = connInfo.gidIdx;
```
AtlantaPepsi, line 2709: I believe this should be the local instead of remote GID. Assuming config file is consistent across all ranks as we do, it shouldn't make a difference. In future we should enforce better gid selection logic like in[ rocSHMEM](https://github.com/ROCm/rocSHMEM/blob/bbad1d8539f442beee893ba0755caabcded71e0e/src/gda/backend_gda.cpp#L1332), especially for RoCE systems

AtlantaPepsi (approved):

src/header/TransferBench.hpp
gilbertlee-amd: Currently we allow for different GIDs per rank.  For now we'll stick with using the remote GID.

PR #232, opened 2026-02-09
DMABUF support in TransferBench

paklui:
## Motivation

This change is to enable DMABUF support in TransferBench. There is some OS kernel with an inbox network driver that only supports dmabuf, and not have the patched peermem support from the out-of-box driver software, therefore we would like TransferBench to also support dmabuf which would allow inbox network driver to work for those kernel. 

## Technical Details

DMA-BUF support requires both the ibverbs API (`ibv_reg_dmabuf_mr`) and ROCm DMA-BUF export API `hsa_amd_portable_export_dmabuf` to function correctly. We should keep existing default for using peermem support, and only enable when `make DISABLE_DMABUF=0` is set at build time. I added some descriptions about DMABUF related API is found in the related headers in `Makefile` so the `HAVE_DMABUF_SUPPORT` is triggered when both APIs are available in check.

## Test Plan
(edited: instead of USE_DMABUF=1 to enable which was used initially, during code review it was suggested to use DISABLE_DMABUF instead, thus edited to use DISABLE_DMABUF=0, to enable)

- Build with DISABLE_DMABUF=1 (default when DISABLE_DMABUF not specified) - it uses legacy peermem support with the message "Building with DMA-BUF support disabled" at build time.
- Build with DISABLE_DMABUF=0 on system missing API (e.g. ibv_reg_dmabuf_mr and/or hsa_amd_portable_export_dmabuf). It will not build with dmabuf support.
- Build with DISABLE_DMABUF=0 - it uses dmabuf support with the message at ""Building with DMA-BUF support". It would use DMABUF support at runtime if the kernel has the `CONFIG_DMABUF_MOVE_NOTIFY=y` and `CONFIG_PCI_P2PDMA=y` 

Tested with CentOS 9 using the `6.16.1-0_fbk1_npi_rc0_brcmrdma2_324_ga601bc980ed0`.
Tested with CentOS 9 using the `6.9.0-0_fbk6_brcmrdma11_125_gfecec9d12677`.
Tested with Ubuntu 24.04.3 LTS with `6.8.0-53-generic`.
Verify RDMA operations work correctly with and without DMA-BUF support
Tested on system with ROCm 7.1.1 with Broadcom Thor2 Ethernet NICs, and tested previously separately on Mellanox ConnectX-7 on one other external config.

## Test Result

Confirmed default build (DISABLE_DMABUF=1) completes successfully
Confirmed DISABLE_DMABUF=0 correctly detects and enables both APIs on 6.16 kernel that only has inbox driver with DMABUF support without peermem support works. 
Verified RDMA transfers work with DMA-BUF support enabled, using `a2a_n` present on single node, and `nicrings` preset on multi-node.

## Submission Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,9 @@ CUDA_PATH ?= /usr/local/cuda
 MPI_PATH  ?= /usr/local/openmpi
 
 # Optional features (set to 0 to disable, 1 to enable)
-# DISABLE_NIC_EXEC: Disable RDMA/NIC executor support
-# DISABLE_MPI_COMM: Disable MPI communicator support
-# USE_DMABUF: Enable DMA-BUF support for GPU Direct RDMA (default: 0)
+# DISABLE_NIC_EXEC: Disable RDMA/NIC executor support (default: 0)
+# DISABLE_MPI_COMM: Disable MPI communicator support (default: 0)
+# DISABLE_DMABUF: Disable DMA-BUF support for GPU Direct RDMA (default: 1)
 
 HIPCC ?= $(ROCM_PATH)/bin/amdclang++
 NVCC ?= $(CUDA_PATH)/bin/nvcc
@@ -93,9 +93,9 @@ ifeq ($(filter clean,$(MAKECMDGOALS)),)
       LDFLAGS += -libverbs
       NIC_ENABLED = 1
 
-      # Disable DMA-BUF support by default (set USE_DMABUF=1 to enable)
-      USE_DMABUF ?= 0
-      ifneq ($(USE_DMABUF), 0)
+      # Disable DMA-BUF support by default (set DISABLE_DMABUF=0 to enable)
+      DISABLE_DMABUF ?= 1
+      ifeq ($(DISABLE_DMABUF), 0)
         # Check for both ibv_reg_dmabuf_mr and ROCm DMA-BUF export support
         HAVE_IBV_DMABUF := $(shell echo '#include <infiniband/verbs.h>' | $(CXX) -E - 2>/dev/null | grep -c 'ibv_reg_dmabuf_mr')
         HAVE_ROCM_DMABUF := $(shell echo '#include <hsa/hsa_ext_amd.h>' | $(CXX) -I$(ROCM_PATH)/include -E - 2>/dev/null | grep -c 'hsa_amd_portable_export_dmabuf')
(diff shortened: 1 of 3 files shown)


src/header/TransferBench.hpp
```diff
@@ -1431,6 +1432,83 @@ namespace {
...
+    }
+
+    bool hasMoveNotify = false;
+    bool hasP2pDma = false;
+
+    // Read kernel config from /boot/config-$(uname -r)
+    struct utsname utsname;
+    if (uname(&utsname) == 0) {
+      char kernel_conf_file[128];
+      snprintf(kernel_conf_file, sizeof(kernel_conf_file), "/boot/config-%s", utsname.release);
```
nileshnegi: @paklui maybe we should add additional paths here (like in https://github.com/ROCm/rocm-systems/blob/a5011ed316de3acd42ec4d968017b5bc75910aca/projects/rccl/src/misc/rocmwrap.cc#L194-L202), in case this file does not exist -- e.g. in the case of VM-based setups

Makefile
```diff
@@ -7,6 +7,11 @@ ROCM_PATH ?= /opt/rocm
 CUDA_PATH ?= /usr/local/cuda
 MPI_PATH  ?= /usr/local/openmpi
 
+# Optional features (set to 0 to disable, 1 to enable)
+# DISABLE_NIC_EXEC: Disable RDMA/NIC executor support
+# DISABLE_MPI_COMM: Disable MPI communicator support
+# USE_DMABUF: Enable DMA-BUF support for GPU Direct RDMA (default: 0)
```
gilbertlee-amd: Is there a reason to not enable DMA-BUF by default in the Makefile, since there is already a fall-back codepath?

If so, this can be set to DISABLE_DMA_BUF to be consistent with the other compile-time features.

gilbertlee-amd (changes requested):
Will also require changes in CMakefile

Makefile
paklui: I think maybe at this time we should keep TransferBench to run using in the way for other test tools, such as perftest, RCCL, UCX, which still uses peermem aas default, and not having DMABUF support enabled by default. We run different test tools to observe difference in behavior usually.

paklui:
> Will also require changes in CMakefile

sure, I made the change also in CMakeList.txt

gilbertlee-amd (approved):

PR #239, opened 2026-03-04
Fix ibv_poll_cq hang due to CQ overflow at high QP

paklui:
## Motivation

After the change, #237 that overcomes the 256 QP limit in TransferBench, I tried to use TransferBench to test the NICs for large QP counts, it is discovered that there is hang would occur (somewhat) randomly around 2020-2030 QPs, the command that generate the hang can be the nicrings or a2a_n presets for NUM_QUEUE_PAIRS=2048. From the stack trace, the hang appears at duirng ibv_cq_poll when large QP is used:
```
$ TB_VERBOSE=1 NUM_QUEUE_PAIRS=2048 ./TransferBench nicrings
...
[INFO] Transfer 7 SubExec 2047 executed by rank 0 NIC 11 is remote write with 1 chunks
[INFO] Transfer 7 SubExec 2047 chunk 0 local 0x7f3f033e0000 remote 0x7f3ef31e0000 of size 131072

$ watch -d -n 1 pstack $(pgrep TransferBench)
Thread 6 (Thread 0x7ecd4f9db640 (LWP 494118) "TransferBench"):
#0  0x00007f2519a7e3d1 in bnxt_re_poll_one (cq=cq@entry=0x312f890, nwc=nwc@entry=1, wc=0x7ecd4f9d2eb0, resize=resize@entry=0x7ecd4f9d2e1c) at /mnt/GT_NFS/paklui/IBGDA/rdma-core-stable-v57/providers/bnxt_re/main.h:550
#1  0x00007f2519a80bea in bnxt_re_poll_cq (ibvcq=0x312f890, nwc=1, wc=0x7ecd4f9d2eb0) at /mnt/GT_NFS/paklui/IBGDA/rdma-core-stable-v57/providers/bnxt_re/verbs.c:1069
#2  0x00000000004ad0bd in TransferBench::(anonymous namespace)::RunExecutor(int, TransferBench::ConfigOptions const&, TransferBench::ExeDevice const&, TransferBench::(anonymous namespace)::ExeInfo&) ()
...
```

Checking a few pstack in loop to watch for progress but it does not change and constantly stuck inside the ib_poll_cq call. The ExecuteNicTransfer would loop over the list of QPs and post WR. if you keep posting without polling, CQ entries accumulate and can exhaust CQ capacity. This may overflow the CQ depth and stall completion accounting. =>which seems to have caused hang when we are having large QPs at around 2020-2030 QPs.

The intent for this test is to try to run with high QP counts to simulate resource consumption on the NIC and driver at larger connection counts or at scale. While trying to look for the cause, the assumption is that the issue (is very likely) due to the current logic that it posts a large number of signaled WRs before polling happens. 

The proposed change to avoid hang at large QP is for the NIC execution path to post and poll in batches (it appears that currently the ibv_poll_cq checks for 1 instead of batches https://github.com/ROCm/TransferBench/blob/dbf9bc1212fdfa6845122bc522751dd7d78c3eea/src/header/TransferBench.hpp#3977

This motivation for this change is to have during NIC transfers to signal periodically and poll while posting to keep the ongoing completions bounded replace unbounded post-all then poll behavior that could hang/stall with large QP counts keep each transfer timing in RunNicExecutor after the new bounded execution path add PollNicCompletion helper function to drain CQEs in batches.

## Technical Details

Previously how it worked (or worked at low QP count)
- After posting all sends for all transfers, https://github.com/ROCm/TransferBench/blob/dbf9bc1212fdfa6845122bc522751dd7d78c3eea/src/header/TransferBench.hpp#L3963-L3967 it starts polling completions. https://github.com/ROCm/TransferBench/blob/dbf9bc1212fdfa6845122bc522751dd7d78c3eea/src/header/TransferBench.hpp#L3968-L3996
- It tracks receivedQPs[i] per transfer. 
- For each transfer still in progress, it polls exactly one CQ entry (ibv_poll_cq(..., 1, &wc)). https://github.com/ROCm/TransferBench/blob/dbf9bc1212fdfa6845122bc522751dd7d78c3eea/src/header/TransferBench.hpp#L3977
- If a CQE arrives and status is good, receivedQPs[i]++.
- When receivedQPs[i] == rss.qpCount, that transfer is considered done and timed.
- It repeats until for all the transfers have been completed in that while loop.
- It assumes “one completion per QP” for completion accounting, as it came from WR setup where only the last chunk in a QP chain was signaled (IBV_SEND_SIGNALED), so polling expected one CQE per QP completion event.

Description for the change
- it changes when/how often completions are requested and polled, changed from 1 to 4 (cqMaxPollEvent). Currently RCCL uses either 4 or 16 depends on the netib patch used. (for Transferbench, an new env allows to it to be adjustable: NIC_CQ_POLL_MAX_EVENT) https://github.com/ROCm/rccl/blob/57e58688f44c77076ad536ef1f6b68741fc6e694/src/transport/net_ib_rocm.cc#L2933-L2971
- it signals periodically, then polls in bounded fashion during posting, instead of post all of the sends, then poll later
- this change should only affect the signaling and notification frequency, and should not change other areas
- comparing to previous, this change creates polling windows, and should lower completion queue pressure, which before it caused CQ overrun/stall that believe that contributes to the hang at large QP count
- the current change still checks every polled Work Completion status and drains required completions before return
- https://github.com/ROCm/TransferBench/compare/develop...paklui:TransferBench:2KQP?expand=1#diff-5a69eadf0bdee9073c96d48a70429756616bb904658b37439878f6a90fd7b234R3975-R3982
- separate out the previous poll for completion into its helper function PollNicCompletions

## Test Plan

Tried to run the failing case for QPs at 2048, 4096, and higher, appeared to work:
```
TB_VERBOSE=1 NUM_QUEUE_PAIRS=2048 ./TransferBench a2a_n
```

## Test Result

It does not hang at large QP used, just take a bit longer to run.

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/src/header/TransferBench.hpp
+++ b/src/header/TransferBench.hpp
@@ -3970,7 +3970,8 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
         bool const signalChunk = isLastChunk || (chunksSinceSignal + 1 >= signalInterval);
 
         ibv_send_wr& wr = rss.sendWorkRequests[qpIndex][chunkIdx];
-        wr.send_flags = signalChunk ? IBV_SEND_SIGNALED : 0;
+        wr.send_flags &= ~IBV_SEND_SIGNALED;
+        if (signalChunk) wr.send_flags |= IBV_SEND_SIGNALED;
 
         int error = ibv_post_send(queuePairs[qpIndex], &wr, &badWorkReq);
         if (error)
--- a/src/header/TransferBench.hpp
+++ b/src/header/TransferBench.hpp
@@ -3932,80 +3932,29 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
   }
 
 #ifdef NIC_EXEC_ENABLED
-  static ErrResult PollNicCompletions(ibv_cq* cq, size_t numCompletions, int transferIdx, int cqMaxPollEvent)
+  // Post one queue pair worth of work requests for a NIC transfer.
+  // Signaling policy is prepared ahead of time in sendWorkRequests: only the
+  // final chunk of each QP is signaled, so one posted QP maps to one CQE.
+  static ErrResult ExecuteNicTransfer(TransferResources&   rss,
+                                      uint32_t      const  qpIndex)
   {
-    int pollBatch = std::max(1, cqMaxPollEvent);
-    std::vector<ibv_wc> wc((size_t)pollBatch);
-    size_t completed = 0;
-
-    while (completed < numCompletions) {
-      int pollCount = (int)std::min(numCompletions - completed, (size_t)pollBatch);
-      int nc = ibv_poll_cq(cq, pollCount, wc.data());
-      if (nc < 0)
-        return {ERR_FATAL, "Transfer %d: Received negative work completion", transferIdx};
-      if (nc == 0)
-        continue;
-
-      for (int idx = 0; idx < nc; idx++) {
-        if (wc[idx].status != IBV_WC_SUCCESS)
-          return {ERR_FATAL, "Transfer %d: Received unsuccessful work completion [status code %d]", transferIdx, wc[idx].status};
-      }
-      completed += nc;
-    }
-
-    return ERR_NONE;
-  }
-
-  // Execution of a single NIC Transfer
-  static ErrResult ExecuteNicTransfer(int           const  iteration,
-                                      ConfigOptions const& cfg,
-                                      int           const  exeIndex,
-                                      TransferResources&   rss)
-  {
-    auto* completionQueue = rss.srcIsExeNic ? rss.srcCompQueue : rss.dstCompQueue;
+    ibv_send_wr* badWorkReq = nullptr;
     auto& queuePairs = rss.srcIsExeNic ? rss.srcQueuePairs : rss.dstQueuePairs;
-
-    size_t const signalInterval = std::max((size_t)1, (size_t)cfg.nic.maxSendWorkReq - 1);
-    size_t const maxOutstandingCompletions = std::max((size_t)1, (size_t)cfg.nic.queueSize / 2);
-
-    // Loop over each queue pair and post/poll in bounded batches to avoid SQ/CQ overflow
-    for (int qpIndex = 0; qpIndex < rss.qpCount; qpIndex++) {
-      ibv_send_wr* badWorkReq = nullptr;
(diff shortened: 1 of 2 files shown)


src/header/TransferBench.hpp
```diff
@@ -3917,24 +3922,75 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
...
+      size_t outstandingCompletions = 0;
+
       for (size_t chunkIdx = 0; chunkIdx < numChunks; chunkIdx++) {
-        int error = ibv_post_send(rss.srcIsExeNic ? rss.srcQueuePairs[qpIndex] : rss.dstQueuePairs[qpIndex],
-                                  &rss.sendWorkRequests[qpIndex][chunkIdx], &badWorkReq);
+        bool const isLastChunk = (chunkIdx + 1 == numChunks);
+        bool const signalChunk = isLastChunk || (chunksSinceSignal + 1 >= signalInterval);
+
+        ibv_send_wr& wr = rss.sendWorkRequests[qpIndex][chunkIdx];
+        wr.send_flags = signalChunk ? IBV_SEND_SIGNALED : 0;
```
mustafabar: may clobber existing flags

```suggestion
        wr.send_flags &= ~IBV_SEND_SIGNALED;
        if (signalChunk) wr.send_flags |= IBV_SEND_SIGNALED;
```

src/header/TransferBench.hpp
```diff
@@ -3917,24 +3922,75 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
...
+        if (signalChunk)
+          outstandingCompletions++;
+
+        if (outstandingCompletions >= maxOutstandingCompletions) {
+          ERR_CHECK(PollNicCompletions(completionQueue, 1, rss.transferIdx, cfg.nic.cqMaxPollEvent));
+          outstandingCompletions--;
+        }
       }
+
+      if (outstandingCompletions)
```
mustafabar: `ExecuteNicTransfer()` now posts and drains completions for one transfer before moving to the next, which appears to serialize transfers that were previously allowed to be in flight together. That changes execution semantics and may reduce overlap / benchmark throughput. Can you comment on the performance aspect of this approach?

paklui: Good point. I agree this can be an issue. I will rework on this and withdraw this PR and would to need rework on it.

PR #256, opened 2026-04-15
Modifying the gfxsweep preset

gilbertlee-amd:
## Motivation
Tweak the `gfxsweep` preset to make it easier to import into spreadsheets / track which set of options yields the best results.
Also allow the use of multiple Transfers / wildcards 

## Technical Details
Re-wrote parts of preset

## Test Result
Here's example output:

```
[GFX Sweep Related]
BLOCKSIZES           =            4 : 256,512,768,1024
GFX_TRANSFER         = R0G0->R0G0->R0G0 : GFX Transfer to sweep (see config file format)
NUM_SUB_EXECS        =            5 : 4,8,16,32,64
TEMPORAL_MODES       =            1 : 0
UNROLLS              =            4 : 1,2,4,8
WAVE_ORDERS          =            1 : 0
WORDSIZES            =            1 : 4
VERBOSE              =            0 : Display summary only

GFX sweep: (1048576 bytes per Transfer). All values are CPU-timed GB/s
=======================================================================================
Transfer     0: (G0->G0->G0)
=======================================================================================
 WvO   WSz   TpM   BlkS   UnR   SE 004  SE 008  SE 016  SE 032  SE 064
  0     4     0     256     1     5.38    6.00    7.79    6.76    4.22
  0     4     0     256     2    11.81    9.23    9.50    6.84    4.64
  0     4     0     256     4    12.05   11.29    8.98    6.44    4.58
  0     4     0     256     8    12.01   11.25    9.52    6.48    4.59
  0     4     0     512     1    11.75   10.65    9.33    6.80    4.63
  0     4     0     512     2    12.27   10.95    9.70    6.66    4.59
  0     4     0     512     4    11.76   11.26    8.54    6.62    4.58
  0     4     0     512     8    10.66   11.44    9.46    7.01    4.51
  0     4     0     768     1    10.09   11.37    8.69    6.91    4.65
  0     4     0     768     2    12.07   11.38    9.03    6.81    4.63
  0     4     0     768     4    12.10   11.30    8.97    6.40    4.66
  0     4     0     768     8    12.33   11.28    9.38    7.13    4.52
  0     4     0    1024     1    12.07   10.96    9.37    7.05    4.62
  0     4     0    1024     2    12.82   10.94    9.34    6.81    4.68
  0     4     0    1024     4    12.11   10.72    9.61    6.28    4.65
  0     4     0    1024     8    11.58   10.82    9.17    6.73    4.46
=======================================================================================
Highest bandwidth found:   12.82 GB/s (CPU-timed)
          WaveOrder    :       0
          WordSize     :       4
          Temporal Mode:       0
          BlockSize    :    1024
          Unroll       :       2
          NumSubExec   :       4


```

### DIFF
--- a/src/client/Presets/GfxSweep.hpp
+++ b/src/client/Presets/GfxSweep.hpp
@@ -135,7 +135,8 @@ int GfxSweepPreset(EnvVars&          ev,
   Utils::Print("          WordSize     : %7d\n", best[1]);
   Utils::Print("          Temporal Mode: %7d\n", best[2]);
   Utils::Print("          BlockSize    : %7d\n", best[3]);
-  Utils::Print("          NumSubExec   : %7d\n", best[4]);
-  Utils::Print("          Unroll       : %7d\n", best[5]);
+  Utils::Print("          Unroll       : %7d\n", best[4]);
+  Utils::Print("          NumSubExec   : %7d\n", best[5]);
+
   return 0;
 }
--- a/src/client/Presets/GfxSweep.hpp
+++ b/src/client/Presets/GfxSweep.hpp
@@ -28,14 +28,14 @@ int GfxSweepPreset(EnvVars&          ev,
                    bool        const bytesSpecified)
 {
   // Collect environment variables for this preset
-  vector<int> blockList     = EnvVars::GetEnvVarArray("BLOCKSIZES",   {256,512,768,1024});
-  std::string transferStr   = EnvVars::GetEnvVar(     "GFX_TRANSFER", "R0G0->R0G0->R0G0");
-  vector<int> numSesList    = EnvVars::GetEnvVarArray("NUM_SUB_EXECS",    {4,8,16,32,64});
-  vector<int> temporalList  = EnvVars::GetEnvVarArray("TEMPORAL_MODES",              {0});
-  vector<int> unrollList    = EnvVars::GetEnvVarArray("UNROLLS",               {1,2,4,8});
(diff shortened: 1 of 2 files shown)


src/client/Presets/GfxSweep.hpp
```diff
@@ -22,169 +22,121 @@ THE SOFTWARE.
...
-                } else {
-                  ++testNum;
-                  TransferBench::Utils::Print("No results available for this sweep point (transfer run failed).\n");
-                }
+                Utils::Print("\n");
+                Utils::PrintErrors(result.errResults);
+                return 1;
               }
             }
+            Utils::Print("\n");
```
alex-breslow-amd, line 124: I saw that you are manually flushing elsewhere.  Maybe it makes sense to fflush here too.  If there's a way to get the raw output stream, then using the << operator with std::endl would work.  You wouldn't have to separately call fflush.  Maybe Print should have this as an optional argument (flush/no flush).  I don't know if it matters too much one way or the other.

alex-breslow-amd (approved):

AtlantaPepsi (approved):

src/client/Presets/GfxSweep.hpp
gilbertlee-amd: Right - this was just to flush any immediate results to the screen.  It likely wouldn't matter much in this case because we'd flush on next result we printed, but it doesn't hurt.

PR #241, opened 2026-03-10
cuda + MNNVL update & pod presets

AtlantaPepsi:

PR #278, opened 2026-04-29
Fixes for cuMem compilation and invalid device ordinal

AtlantaPepsi:
## Motivation
cuMem symbols and definitions are currently guarded by pod communication enablement. It's not a long term solution as these two are not always coupled and usage might diverge in future. The separation of these two also fixes existing linking error with `cuMemcpyAsync` or `CUResult` in absence of `POD_COMM_ENABLED`.

## Technical Details

- Introduced a separate `CUMEM_ENABLED` macro for build process as well as header. `POD_COMM_ENABLED` will have to depend on cuMem enablement as well. 
- Minor adjustment for error reporting for cuda runtime calls
- Eliminated unnecessary `hipSetDevice`: Previously in #241 multiple `hipSetDevice` are added throughout cuMem allocation and release to make sure context is always initialized. Not all of them were needed, and unconditional invocation also caused error for non GPU executors.

## Test Plan
Tested all combination of Makefile flags and made sure compilation/linking succeeded.
Previously on CI machines with more CPU NUMA nodes than GPU devices, certain sweeping presets such as p2p would fail, which is fixed now.

## Submission Checklist

### DIFF
--- a/src/header/TransferBench.hpp
+++ b/src/header/TransferBench.hpp
@@ -5453,8 +5453,9 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
         Transfer const& t = transfers[resource->transferIdx];
         for (int srcIdx = 0; srcIdx < resource->srcMem.size(); srcIdx++) {
           if (t.srcs[srcIdx].memRank == localRank) {
-            if (IsGpuMemType(t.srcs[srcIdx].memType))
+            if (IsGpuMemType(t.srcs[srcIdx].memType)) {
               ERR_APPEND(hipSetDevice(t.srcs[srcIdx].memIndex), errResults);
+            }
             ERR_APPEND(hipMemcpy(resource->srcMem[srcIdx] + initOffset, srcReference[srcIdx].data(), resource->numBytes,
                                  hipMemcpyDefault), errResults);
           }
--- a/src/header/TransferBench.hpp
+++ b/src/header/TransferBench.hpp
@@ -1460,8 +1460,6 @@ namespace {
     // If memHandle is provided, allocate sharable memory
     if (memHandle != NULL) {
 #ifdef POD_COMM_ENABLED
-      ERR_CHECK(hipSetDevice(deviceIdx));
-      // Prepare HIP memory allocation properties structure
       hipMemAllocationProp prop = {};
       ERR_CHECK(GetMemAllocationProp(memDevice, prop));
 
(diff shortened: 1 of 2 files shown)


src/header/TransferBench.hpp
```diff
@@ -5455,7 +5453,8 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
         Transfer const& t = transfers[resource->transferIdx];
         for (int srcIdx = 0; srcIdx < resource->srcMem.size(); srcIdx++) {
           if (t.srcs[srcIdx].memRank == localRank) {
-            ERR_APPEND(hipSetDevice(t.srcs[srcIdx].memIndex), errResults);
+            if (IsGpuMemType(t.srcs[srcIdx].memType))
+              ERR_APPEND(hipSetDevice(t.srcs[srcIdx].memIndex), errResults);
```
nileshnegi:
```suggestion
            if (IsGpuMemType(t.srcs[srcIdx].memType)) {
              ERR_APPEND(hipSetDevice(t.srcs[srcIdx].memIndex), errResults);
            }
```

nileshnegi (approved):

src/header/TransferBench.hpp
nileshnegi: minor nit for readability and TxB's coding format

src/header/TransferBench.hpp
```diff
@@ -4276,13 +4274,11 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
       // Deallocate source memory
       for (int iSrc = 0; iSrc < t.srcs.size(); ++iSrc) {
         if (t.srcs[iSrc].memRank == localRank) {
-          ERR_CHECK(hipSetDevice(t.srcs[iSrc].memIndex));
```
alex-breslow-amd, line 4279: QQ: How come we no longer have to set the hip device?  Nvm, I see that you say that some are redundant.  Should we at least check that the device that we're on is the correct one?

src/header/TransferBench.hpp
```diff
@@ -5312,36 +5306,40 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
...
       this->errMsg  = "";
     } else {
-      const char *errString = NULL;
-      hsa_status_string(err, &errString);
+      const char *errString = NULL, *errName = NULL;
+      cuGetErrorName(err, &errName);
+      cuGetErrorString(err, &errString);
       this->errType = ERR_FATAL;
-      this->errMsg  = std::string("HSA Error: ") + errString;
+      this->errMsg  = std::string("CUDA Driver Error: ") + errName
```
alex-breslow-amd, line 5341: I don't think fmt is an existing dependency of TransferBench, but this would be cleaner with that package.

src/header/TransferBench.hpp
```diff
@@ -5312,36 +5306,40 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
       this->errMsg  = "";
     } else {
       this->errType = ERR_FATAL;
+#if defined(__NVCC__)
+      this->errMsg  = std::string("CUDA Runtime Error: ") + hipGetErrorString(err);
```
alex-breslow-amd, line 5310: Shouldn't this be cudaGetErrorString?  Or are we using hip on NVIDIA hardware too?  Nvm, I think Gilbert mentioned there are set of defines in TransferBench where the calls get replaced by cuda calls.

PR #284, opened 2026-05-03
Fixing missing std::max, updating client description to include POD support

gilbertlee-amd:
## Motivation
There was a `max` instead of a `std::max` which would cause compilation issues on some setups. 
Additionally, the TransferBench version string has now been updated to include whether or not POD support has been enabled at compile time.


src/client/Client.cpp
nileshnegi, line 26:
```suggestion
#include <algorithm>
#include <fstream>
```

nileshnegi (approved):

PR #286, opened 2026-05-03
Small fixes for candidate v1.67 release

nileshnegi:
## Motivation

Tighten preset input validation, fix correctness bugs, and improve code robustness across several presets and shared utilities.

## Technical Details

Preset validation fixes:
- PodAllToAll:
  - restrict to single-pod runs; multi-pod support is not yet implemented (rejects when GetRankPerPodMap().size() != 1, with a dedicated message for the no-pods-detected case)
  - validate GROUP_SIZE divisibility against the single pod's total device count
  - fix GROUP_SIZE default to use numGpus (user-specified) instead of numDetectedGpus to avoid false validation failures when NUM_GPU_DEVICES is set below the detected count
  - guard GROUP_SIZE < 1 before modulo to prevent UB
  - guard numGpus <= 0 (was only catching < 0)
- WallClock:
  - guard numXccs <= 0 before kernel launch to prevent invalid gridDim.x=0 dispatch when topology data is missing
  - zero-initialize timestamps buffer before each kernel launch to prevent stale values from unexpected XCC IDs corrupting delta calculations
  - reject NUM_ITERATIONS <= 0 (seconds-based and infinite modes are unsupported)
  - extend NUM_GPU_DEVICES validation to also reject values exceeding detected GPU count
  - use sizeof(*readyFlag) for type safety
- NicAllToAll:
  - guard bwByRankNic indexing against remapped NIC executor indices from RunTransfers
  - emit a warning instead of silently skipping on out-of-range results

Code quality:
- EnvVars:
  - extract shared GetEnvTokens helper to eliminate duplicated getenv/strtok logic across three array-parsing functions
  - treat unset/empty env vars consistently as "use default"
  - fix GetCuMaskDesc dropping the last contiguous CU run
  - use size_t loop index in GetStr to fix signed/unsigned comparison
- Add missing explicit STL includes that were previously resolved only through transient HIP/compiler headers

## Submission Checklist

### DIFF
--- a/src/client/Presets/PodAllToAll.hpp
+++ b/src/client/Presets/PodAllToAll.hpp
@@ -43,6 +43,11 @@ int PodAllToAllPreset(EnvVars&          ev,
   int numRanks = TransferBench::GetNumRanks();
   int numDetectedGpus = TransferBench::GetNumExecutors(EXE_GPU_GFX);
 
+  // When pod detection fails (e.g. amd-smi unavailable), the map is empty
+  if (Utils::GetRankPerPodMap().empty()) {
+    Utils::Print("[ERROR] No pods detected. Set TB_FORCE_SINGLE_POD=1 to treat all ranks as a single pod.\n");
+    return ERR_FATAL;
+  }
   // Restrict to single-pod runs; multi-pod support is not yet implemented
   if (Utils::GetRankPerPodMap().size() != 1) {
     Utils::Print("[ERROR] PodAllToAll preset currently requires all ranks to be in a single pod. Set TB_FORCE_SINGLE_POD=1 to treat all ranks as a single pod.\n");
@@ -129,7 +134,7 @@ int PodAllToAllPreset(EnvVars&          ev,
     return ERR_FATAL;
   }
 
-  // Validate per-pod: pods may have different rank counts so global check is insufficient
+  // Validate group size divides the single pod's total device count
   for (auto const& [pod, ranks] : Utils::GetRankPerPodMap()) {
     int podDevices = (int)ranks.size() * numGpus;
     if (podDevices % groupSize) {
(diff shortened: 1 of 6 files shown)


gilbertlee-amd (approved):

src/client/Presets/PodAllToAll.hpp
```diff
@@ -118,11 +118,21 @@ int PodAllToAllPreset(EnvVars&          ev,
...
   }
 
-  if (numRanks * numDetectedGpus % groupSize) {
-    Utils::Print("[ERROR] Group size %d cannot evenly divide %d total devices from %d ranks.\n", groupSize, numRanks * numDetectedGpus, numRanks);
+  if (groupSize < 1) {
+    Utils::Print("[ERROR] GROUP_SIZE must be >= 1 (got %d)\n", groupSize);
     return ERR_FATAL;
   }
 
+  // Validate per-pod: pods may have different rank counts so global check is insufficient
```
AtlantaPepsi: For now we are limiting to single pod

AtlantaPepsi (dismissed):

src/client/Presets/PodAllToAll.hpp
nileshnegi: yeah, but this check would be innocent, right?
only put it cause Copilot complained about it several times on different PRs :p

AtlantaPepsi: We should enforce single pod by restricting ` Utils::GetRankPerPodMap()` to size 1, in which case this condition would be trivial. 
When we extend support to multipod in future, I also wonder if a uniform group size for all pods would make sense. For now I think it's best we don't add per pod group size check

nileshnegi: then change the code limitations to be for 1 pod only.
i see this as a trivial check, even if we run 1 pod, and don't see the need to remove it. how does it deter this preset?

nileshnegi: if you say ERR_FATAL is wrong, and ERR_WARN is better to allow expts., sure we can go for that.
but i want a check.

gilbertlee-amd: Yes.  Please just add in the check for Utils::GetRankPerPodMap().size() == 1.

gilbertlee-amd (dismissed):

AtlantaPepsi (approved):

PR #262, opened 2026-04-23
CI: build relocatable packages via TheRock SDK

thananon:
## Summary

Adds GitHub Actions workflow and build script for producing relocatable packages (DEB/RPM/TGZ) via the TheRock SDK. This workflow is the primary path for continuous integration and nightly/release builds of TransferBench packages, enabling self-contained deployments without requiring a pre-installed ROCm stack on the target machine.

**Prerequisites:** Merged #261 (manylinux portability fixes) to ensure builds succeed on manylinux_2_28 (AlmaLinux 8) with gcc 8-era libstdc++.

## Implementation

- **`.github/workflows/build-relocatable-packages.yml`**: Two-job workflow (Ubuntu 22.04 → DEB/TGZ, manylinux_2_28 → RPM/TGZ) that fetches the latest TheRock SDK tarball, builds with `-DBUILD_RELOCATABLE_PACKAGE=ON`, and uploads artifacts to GitHub Actions + optional S3 (via OIDC).
- **`build_packages_local.sh`**: Standalone script that downloads TheRock SDK tarballs from S3 by ROCm version and GPU family (gfx94X-dcgpu, gfx950-dcgpu, gfx110X-all, gfx120X-all, gfx1151), extracts them, drives CMake with relocatable RPATH, and produces `amdrocm<MAJOR>-transferbench` packages.
- **CMake additions**: New `BUILD_RELOCATABLE_PACKAGE` option that bypasses `rocm_install`/`rocm_create_package` and drives CPack directly with version-namespaced package naming (`amdrocm7-transferbench`) and relocatable RPATH.

## S3 and OIDC

Workflow uploads packages to S3 using GitHub OIDC for authentication (no long-lived credentials). S3 paths:
- **Nightly/develop**: `s3://.../nightly/transferbench/{deb,rpm,tar}` with apt/yum repo metadata.
- **Release branches**: `s3://.../release/transferbench/{deb,rpm,tar}` with apt/yum metadata.
- **Pull requests**: `s3://.../transferbench/<branch>/<run-number>/{ubuntu-22.04,manylinux_2_28}` (no metadata).

S3 upload only executes in the `ROCm/TransferBench` repository and requires `AWS_S3_BUCKET` (organization variable) and `AWS_ROLE_ARN` (organization secret).

## Supported GPU Families

TheRock SDK tarballs are fetched per GPU family:
- **gfx94X-dcgpu**: gfx940, gfx941, gfx942, gfx943 (MI300 series)
[...]

PR #287, opened 2026-05-04
ci: align relocatable package flow with RVS reference

thananon:
## Summary

Brings TransferBench's TheRock-SDK-based package build into closer parity with the [ROCmValidationSuite reference flow](https://github.com/ROCm/ROCmValidationSuite/blob/master/.github/workflows/README_BUILD_PACKAGES.md). Closes the four gaps identified after #262 landed.

### Gap 1 — RPATH defaults move into `CMakeLists.txt`
RPATH list previously lived only in `build_packages_local.sh` (`-DCMAKE_INSTALL_RPATH=…`), so plain `cmake -DBUILD_RELOCATABLE_PACKAGE=ON ..` produced a non-relocatable binary. Defaults now live inside the `if(BUILD_RELOCATABLE_PACKAGE)` block in `CMakeLists.txt`. The script drops the three `-DCMAKE_*_RPATH` flags. The list is also extended to match RVS:
```
$ORIGIN:$ORIGIN/../lib:/opt/rocm/extras-<MAJOR>/lib:/opt/rocm/lib:/opt/rocm/lib/llvm/lib:/opt/rocm/core-<MAJOR>/lib:/opt/rocm/core-<MAJOR>/lib/llvm/lib
```

### Gap 2 — Auto-computed patch version
`VERSION_STRING` patch is now derived from `git describe --tags --abbrev=0 --match v<MAJOR>.<MINOR>.*` followed by `git rev-list --count <tag>..HEAD`. Falls back to a hardcoded patch (`02`) when git is unavailable or no matching tag exists. **No user-visible change today** since no `v1.66.*` tag exists yet (latest v-tag is `v1.62.00`); once a `v1.66.00` tag is created, builds will be `1.66.<commits-since>`.

### Gap 3 — RVS-style package release tag
`build_packages_local.sh` now emits release strings matching the RVS format:

| Trigger | Release string |
|---|---|
| push / schedule / dispatch / local | `r<libpatch>.<yyyymmdd>` (e.g. `r0711.20260504`) |
| pull request | `r<libpatch>.<yyyymmdd>.<src-branch>.<commit>` |
| `release/*` branch | `${GITHUB_RUN_NUMBER}` (fallback `1`) |

UTC date captured once at script start so long builds can't straddle midnight. The over-permissive bare `rel*` branch matcher is dropped (workflow only triggers `release/**`).

### Gap 4 — User-facing TGZ install doc
TGZ install instructions move out of the CI-engineer-facing `README_BUILD_PACKAGES.md` into a new Sphinx doc at `docs/install/INSTALL_TGZ.rst`, covering pre-install ROCm, runtime deps (`numactl`, `libnuma1`, `hsa-rocr` — no PCI deps since TransferBench doesn't enumerate via PCI), extraction, `PATH`/`LD_LIBRARY_PATH`, and persistent `/etc/profile.d` setup. `install.rst` cross-links to it. `README_BUILD_PACKAGES.md` keeps a minimal CI-maintainer smoke test plus a link to the new doc.

## Test plan

- [x] `cmake -B build_verify -DBUILD_RELOCATABLE_PACKAGE=ON -DROCM_MAJOR_VERSION=7 …` prints `-- TransferBench version: 1.66.02` (correct fallback)
- [x] Bash release-tag logic smoke-tested in default / release / PR modes — produces `r0711.20260504`, `42`, `r0711.20260504.feature.foo.<commit>` respectively
- [ ] CI: push triggers `build-relocatable-packages.yml`; verify package filenames carry the new `r<libpatch>.<yyyymmdd>` release string
- [ ] CI: confirm `readelf -d` on the built binary shows the extended RPATH list and no `$HOME/rocm-sdk/install` leak
- [ ] Render Sphinx docs (`python3 -m sphinx -W -b html …`) and confirm `INSTALL_TGZ.html` builds without warnings

### DIFF
--- a/.github/workflows/README_BUILD_PACKAGES.md
+++ b/.github/workflows/README_BUILD_PACKAGES.md
@@ -103,8 +103,15 @@ Quick smoke test from the repo root after a successful build:
 sudo mkdir -p /opt/rocm/extras-7
 sudo tar -xzf build/amdrocm7-transferbench-*.tar.gz -C /opt/rocm/extras-7 --strip-components=1
 export PATH=/opt/rocm/extras-7/bin:$PATH
-TransferBench --help
+# With no args, TransferBench prints version, usage, available presets,
+# and detected topology — a fast end-to-end check that the binary loads
+# its ROCm libs correctly.
+TransferBench
 ```
+
+> Once the `help` and `smoketest` presets land on `develop` (currently on
+> `candidate`), prefer `TransferBench help` for usage and `TransferBench
+> smoketest` for a real correctness check.
 
 ## S3 upload (OIDC)
 
--- a/.github/workflows/README_BUILD_PACKAGES.md
+++ b/.github/workflows/README_BUILD_PACKAGES.md
@@ -93,11 +93,17 @@ sudo rpm -i --replacefiles --nodeps build/amdrocm7-transferbench-*.rpm
 
 ### Any Linux (TGZ — relocatable install tree, requires ROCm runtime on target)
 
+End-user instructions (pre-install ROCm, runtime dependencies, extract,
+`PATH` / `LD_LIBRARY_PATH`, troubleshooting) live in the project docs at
+[docs/install/INSTALL_TGZ.rst](../../docs/install/INSTALL_TGZ.rst).
+
+Quick smoke test from the repo root after a successful build:
+
 ```bash
 sudo mkdir -p /opt/rocm/extras-7
 sudo tar -xzf build/amdrocm7-transferbench-*.tar.gz -C /opt/rocm/extras-7 --strip-components=1
 export PATH=/opt/rocm/extras-7/bin:$PATH
-TransferBench
+TransferBench --help
 ```
 
 ## S3 upload (OIDC)
--- a/docs/install/INSTALL_TGZ.rst
+++ b/docs/install/INSTALL_TGZ.rst
@@ -1,0 +1,132 @@
+:orphan:
+
+.. meta::
+  :description: Install the relocatable TransferBench TGZ archive on any Linux distribution
+  :keywords: TransferBench, TGZ, tarball, install, relocatable
+
+.. _install-transferbench-tgz:
+
+------------------------------------------------
+Installing TransferBench from the TGZ archive
+------------------------------------------------
+
+The TransferBench TGZ archive (``amdrocm<MAJOR>-transferbench-*.tar.gz``) is a
+relocatable install tree that works on any Linux distribution where a
+compatible ROCm runtime is already present. Use it when you cannot or do not
+want to install the DEB or RPM package — for example on a distribution
+without a native ROCm package, or inside a non-root container.
+
+The TGZ ships only the ``TransferBench`` binary and its supporting files. It
+does **not** bundle ROCm; the host system must already provide the ROCm
+runtime libraries (``hsa-rocr`` and the HIP runtime).
+
+Pre-install: ROCm
+-----------------
+
+Install ROCm on the target system before extracting the TGZ. Follow the
+official AMD documentation:
+
+* `ROCm documentation <https://rocm.docs.amd.com/>`_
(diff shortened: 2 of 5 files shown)


.github/workflows/README_BUILD_PACKAGES.md
```diff
@@ -93,11 +93,17 @@ sudo rpm -i --replacefiles --nodeps build/amdrocm7-transferbench-*.rpm
...
+[docs/install/INSTALL_TGZ.rst](../../docs/install/INSTALL_TGZ.rst).
+
+Quick smoke test from the repo root after a successful build:
+
 ```bash
 sudo mkdir -p /opt/rocm/extras-7
 sudo tar -xzf build/amdrocm7-transferbench-*.tar.gz -C /opt/rocm/extras-7 --strip-components=1
 export PATH=/opt/rocm/extras-7/bin:$PATH
-TransferBench
+TransferBench --help
```
alex-breslow-amd: Nice.  Is this enough?  It's definitely better than what we have presently, which is zilch.  I'm wondering if we need to run some simple workflows like a2a.

docs/install/INSTALL_TGZ.rst
```diff
@@ -0,0 +1,132 @@
...
+
+  TransferBench --help
+
+You should see the TransferBench usage banner. If the binary fails to load a
+shared library, inspect:
+
+.. code-block:: bash
+
+  ldd /opt/rocm/extras-7/bin/TransferBench
+  readelf -d /opt/rocm/extras-7/bin/TransferBench | grep -E 'RPATH|RUNPATH'
```
alex-breslow-amd, line 119: Should we just run this before running help?  It saves us from having to type this in a terminal.

.github/workflows/README_BUILD_PACKAGES.md
gilbertlee-amd: --help doesn't exist.

In candidate we already introduce the "help" preset, to be more inline with the other presets:

`./TransferBench help`

For CI moving forward, we likely want to use 
`./TransferBench smoketest`

However, that's also in candidate branch still

alex-breslow-amd:
I would try to automate the first-level troubleshooting you listed if possible.  Then the doc can just refer the user to the output that's already part of the logs.

thananon:
@gilbertlee-amd good catch — `--help` was wishful thinking on my part. Pushed a fix that:

- Replaces `TransferBench --help` with plain `TransferBench` in both the README smoke test and `INSTALL_TGZ.rst`. On `develop` today, `argc <= 1` (Client.cpp:41) prints version + usage + preset list + detected topology then exits 0, which is the right load-time check.
- Adds a forward-looking note in the README that once the `help` and `smoketest` presets graduate from `candidate` to `develop`, the smoke test should switch to `TransferBench smoketest`.

Left @alex-breslow-amd's two open suggestions (run a real workload like a2a; auto-run the troubleshooting `ldd`/`readelf`) for follow-up — both are good ideas but bigger than this PR's scope (the a2a one in particular wants the `smoketest` preset to land first).

alex-breslow-amd:
> @gilbertlee-amd good catch — `--help` was wishful thinking on my part. Pushed a fix that:
> 
> * Replaces `TransferBench --help` with plain `TransferBench` in both the README smoke test and `INSTALL_TGZ.rst`. On `develop` today, `argc <= 1` (Client.cpp:41) prints version + usage + preset list + detected topology then exits 0, which is the right load-time check.
> * Adds a forward-looking note in the README that once the `help` and `smoketest` presets graduate from `candidate` to `develop`, the smoke test should switch to `TransferBench smoketest`.
> 
> Left @alex-breslow-amd's two open suggestions (run a real workload like a2a; auto-run the troubleshooting `ldd`/`readelf`) for follow-up — both are good ideas but bigger than this PR's scope (the a2a one in particular wants the `smoketest` preset to land first).

That's fine by me.

thananon:
Codeql broken unrelated to the PR. If we are ok lets merge this and create a test release branch to verify this workflow end to end.

nileshnegi (approved):

PR #292, opened 2026-05-08
revert pinned hostmem for pod comm

AtlantaPepsi:
## Motivation

This is to add back the option of extended GPU memory in cross pod transfer.

## Technical Details

Removed `CheckPages()` inside `AllocateMemory()` for pinned host memory. This would cause a silent fail on the node where it's allocated, and other nodes will hang at broadcast inside `ExchangeMemory`.

It's also not tested yet on either Nvidia or AMD platforms.

## Submission Checklist

### DIFF
--- a/src/header/TransferBench.hpp
+++ b/src/header/TransferBench.hpp
@@ -4094,7 +4094,9 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
       hipError_t exportErr = hipSuccess;
       const char* exportStep = "hipSetDevice";
       if (memDevice.memRank == GetRank()) {
-        exportErr = hipSetDevice(memDevice.memIndex);
+        if (IsGpuMemType(memDevice.memType)) {
+          exportErr = hipSetDevice(memDevice.memIndex);
+        }
         if (exportErr == hipSuccess) {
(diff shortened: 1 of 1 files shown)


src/header/TransferBench.hpp
```diff
@@ -4075,7 +4094,9 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
       hipError_t exportErr = hipSuccess;
       const char* exportStep = "hipSetDevice";
       if (memDevice.memRank == GetRank()) {
-        exportErr = hipSetDevice(memDevice.memIndex);
+        if (IsGpuMemType(memDevice.memType)) {
+          exportErr = hipSetDevice(memDevice.memIndex);
+        }
```
gilbertlee-amd, line 4099: Why is this hipSetDevice even necessary?  The runtime should already have the ability to determine where the memory handle is.  It also seems strange that you would not need it for CPU memory types but GPU memory types.

Based on the documentation, cuMemExportToSharableHandle doesn't require the current device context either, and won't ever return CUDA_ERROR_INVALID_CONTEXT

gilbertlee-amd (changes requested):

PR #295, opened 2026-05-10
Enable no-src GPU SDMA transfers

nileshnegi:
## Motivation

Extends EXE_GPU_DMA to accept zero-source transfers, enabling SDMA-driven memset without introducing a new executor type.

## Technical Details

Fill value:
```
uint32_t fillVal = bit_cast<uint32_t>(MEMSET_VAL);  // 0x4B4B4B4B
count = numBytes / sizeof(uint32_t);  // count is in uint32_t units
```
`0x4B4B4B4B` matches both memset(MEMSET_CHAR) (used by dstReference[0]) and MemsetVal<float>() used by the GFX 0-src kernel, so existing correctness validation passes without changes.

Validation:
- `srcs.size() != 1 → srcs.size() > 1`; 0 sources now valid on AMD
- 0-src + `exeSubIndex != -1` results in fatal error
- Copy-agent-selection warnings wrapped in `if (!t.srcs.empty())`

Resource setup:
- `hsa_amd_pointer_info` on `srcMem[0]` guarded by `!rss.srcMem.empty()`

Constraints:
- AMD only. NVIDIA builds get a fatal error for 0-src DMA transfers.
- No engine selection as `hsa_amd_memory_fill` has no engine parameter/mask, so combining 0 sources with an engine subindex (e.g., `n d0.2 g1`) is rejected at validation.

## Submission Checklist

### DIFF
--- a/src/header/TransferBench.hpp
+++ b/src/header/TransferBench.hpp
@@ -2327,12 +2327,20 @@ namespace {
         }
         break;
       case EXE_GPU_DMA:
-        if (t.srcs.size() != 1) {
+        if (t.srcs.size() > 1) {
           errors.push_back({ERR_FATAL,
-                            "Transfer %d: DMA executor must have exactly 1 source", i});
+                            "Transfer %d: DMA executor must have 0 or 1 sources", i});
           hasFatalError = true;
           break;
         }
+#if defined(__NVCC__)
+        if (t.srcs.empty()) {
+          errors.push_back({ERR_FATAL,
+                            "Transfer %d: DMA memset (0 sources) not supported on NVIDIA hardware", i});
+          hasFatalError = true;
+          break;
+        }
+#endif
         if (t.dsts.size() < 1) {
           errors.push_back({ERR_FATAL,
                             "Transfer %d: DMA executor must have at least 1 destination", i});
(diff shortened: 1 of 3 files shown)


src/header/TransferBench.hpp
```diff
@@ -2341,6 +2349,12 @@ namespace {
           hasFatalError = true;
           break;
 #else
+          if (t.srcs.empty()) {
+            errors.push_back({ERR_FATAL,
+                              "Transfer %d: DMA memset (0 sources) does not support engine selection", i});
```
gilbertlee-amd, line 2368: Unnecessary to list "(0 sources)"

gilbertlee-amd (changes requested):
We likely will need to guard this feature with a run-time driver version check to ensure that it doesn't fall back to a blit kernel

src/header/TransferBench.hpp
```diff
@@ -2313,12 +2313,20 @@ namespace {
...
+          errors.push_back({ERR_FATAL,
+                            "Transfer %d: DMA executor must have 0 or 1 sources", i});
+          hasFatalError = true;
+          break;
+        }
+#if defined(__NVCC__)
+        if (t.srcs.empty()) {
           errors.push_back({ERR_FATAL,
-                            "Transfer %d: DMA executor must have exactly 1 source", i});
+                            "Transfer %d: DMA memset (0 sources) not supported on NVIDIA hardware", i});
```
nileshnegi, line 2339:
```suggestion
                            "Transfer %d: DMA memset not supported on NVIDIA hardware", i});
```

src/header/TransferBench.hpp
nileshnegi, line 2368:
```suggestion
                              "Transfer %d: DMA memset does not support engine selection", i});
```

PR #304, opened 2026-05-19
Fix closest-CPU-NUMA detection regression for GPU executors

nileshnegi:
## Motivation

Restore the Linux-NUMA-indexed cpuAgents[] table that the GPU-to-closest-CPU lookup in CollectTopology() depends on. 

## Technical Details

After #303, CPU agents were enumerated via hsa_iterate_agents + HSA_AGENT_INFO_NODE and stored at the HSA/KFD topology node id, which is not guaranteed to equal the Linux NUMA index (e.g., on systems where GPU agents are enumerated before CPU agents in the KFD topology). When the two diverged, cpuAgents[i] stayed default-initialized, the closestCpuAgent lookup never matched, and GetClosestCpuNumaToGpu() returned -1 for every GPU. That broke the topology display (all GPUs reported "no closest NUMA") and produced malformed transfer descriptors like "R0C-1" in any preset that consumes the result.

Switch back to the allocation-based enumeration that preserves cpuAgents[i] == "agent owning Linux NUMA i" while still tolerating NUMA nodes restricted by the process's cpuset / mempolicy: AllocateMemory(MEM_CPU) already rejects such nodes since #303, so we skip them on ERR_FATAL and leave the slot zero-initialized. Restricted NUMAs cannot service transfers anyway, so a zero-handle entry is the correct sentinel and harmless to the handle comparison performed by the closest-CPU lookup.

Also, drop the explicit hsa_init() / hsa_shut_down() around the iteration: HIP already manages HSA lifetime for this process, and tearing it down between topology collection and the rest of the run is unnecessary.

## Test Result

Verified on 8 x gfx950, 2 NUMA nodes:
- ./TransferBench now prints the expected "NUMA 00 -> GPUs 0 1 2 3" / "NUMA 01 -> GPUs 4 5 6 7" mapping that pre-#303 (e8edacf) produced.

## Submission Checklist

### DIFF
--- a/src/header/TransferBench.hpp
+++ b/src/header/TransferBench.hpp
@@ -7756,39 +7756,36 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
       hsa_amd_pointer_info_t info;
       info.size = sizeof(info);
 
-      // Callback to process each agent
-      auto cpuAgentCallback = [](hsa_agent_t agent, void* data) -> hsa_status_t {
-        std::map<int, hsa_agent_t>* agents = static_cast<std::map<int, hsa_agent_t>*>(data);
-
-        hsa_device_type_t deviceType;
-        hsa_status_t status = hsa_agent_get_info(agent, HSA_AGENT_INFO_DEVICE, &deviceType);
-        if (status != HSA_STATUS_SUCCESS) return status;
-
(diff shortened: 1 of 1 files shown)


nileshnegi:
Fixed in https://github.com/ROCm/TransferBench/pull/305

PR #323, opened 2026-06-11
fix(gfx): copy subExecParam from device before reading timing data

fcui-amd:
## Motivation

The GFX executor crashes with a SIGSEGV when collecting per-Transfer timing
data on GPUs that do not expose VRAM to the host (e.g. PCIe cards without
large/resizable BAR).

## Technical Details

Copy the subExecutor parameters back to host with `hipMemcpy(..., DeviceToHost)`
before reading timing/CU data, in both affected paths. No behavior change on
hardware where direct access previously worked.

## Submission Checklist

### DIFF
--- a/src/header/TransferBench.hpp
+++ b/src/header/TransferBench.hpp
@@ -1478,6 +1478,22 @@ namespace {
     return "?";
   }
 
+  // Returns whether a memory allocation can be directly read by host code.
+  static ErrResult GetMemHostAccessibility(MemDevice const& memDevice, bool& hostAccessible)
+  {
+    hostAccessible = IsCpuMemType(memDevice.memType) || memDevice.memType == MEM_MANAGED;
+    if (hostAccessible || !IsGpuMemType(memDevice.memType)) return ERR_NONE;
+
+#if defined(__NVCC__)
+    return ERR_NONE;
+#else
+    int isLargeBar = 0;
+    ERR_CHECK(hipDeviceGetAttribute(&isLargeBar, hipDeviceAttributeIsLargeBar, memDevice.memIndex));
+    hostAccessible = (isLargeBar != 0);
+    return ERR_NONE;
+#endif
+  }
+
   // Allocate memory
   static ErrResult AllocateMemory(MemDevice memDevice, size_t numBytes, void** memPtr,
                                   size_t* actualBytes = NULL,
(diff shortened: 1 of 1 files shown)


fcui-amd:
ping?

gilbertlee-amd:
This was somewhat intentional, to try to avoid explicit SDMA traffic during the main loop.

The solution for non-AMD hardware was actually to use [managed memory instead] for the GPU parameter buffer.  https://github.com/ROCm/TransferBench/blob/2bc42cd82f239f64168f492794ce0aab770589a4/src/header/TransferBench.hpp#L4371    

I think this PR needs to be more targeted, and only be applied to GPUs that require this functionality.  
Does [isLargeBar](https://rocm.docs.amd.com/projects/HIP/en/7.13.0-preview/doxygen/html/structhip_device_prop__t.html#ad9aa31dd7bd61c84c5fd92d20d0f8b29) report 0 for your device?

fcui-amd:
It's amd gpu with large bar disabled.

src/header/TransferBench.hpp
```diff
@@ -5383,9 +5383,15 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
         if (cfg.general.recordPerIteration) {
           rss.perIterMsec.push_back(deltaMsec);
           std::set<std::pair<int,int>> CUs;
+          // Copy subExecutor parameters back to host before reading them.
+          // subExecParamGpuPtr lives in device memory and is not guaranteed to be host-accessible
+          // (e.g. GPUs without large/resizable BAR), so dereferencing it directly can segfault.
+          std::vector<SubExecParam> subExecParamHost(numSubExecs);
+          ERR_CHECK(hipMemcpy(subExecParamHost.data(), rss.subExecParamGpuPtr,
+                              numSubExecs * sizeof(SubExecParam), hipMemcpyDefault));
```
gilbertlee-amd: This should be gated only for GPUs that do not have largeBar support, so that it doesn't change the behavior of GPUs that do have largeBar support.

src/header/TransferBench.hpp
```diff
@@ -5452,18 +5458,25 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
       // If Transfers were combined into a single launch, figure out per-Transfer timing
       // Determine timing for each of the individual transfers that were part of this launch
       if (!cfg.gfx.useMultiStream) {
+        // Copy subExecutor parameters back to host before reading timing data.
+        // subExecParamGpu lives in device memory and is not guaranteed to be host-accessible
+        // (e.g. GPUs without large/resizable BAR), so dereferencing it directly can segfault.
+        std::vector<SubExecParam> subExecParamHost(exeInfo.totalSubExecs);
+        ERR_CHECK(hipMemcpy(subExecParamHost.data(), exeInfo.subExecParamGpu,
+                            exeInfo.totalSubExecs * sizeof(SubExecParam), hipMemcpyDefault));
+
```
gilbertlee-amd: Same as above.  Gate this copy based on whether or not this GPU has largeBar support.

gilbertlee-amd (changes requested):

src/header/TransferBench.hpp
fcui-amd: updated

gilbertlee-amd (approved):

PR #327, opened 2026-06-28
Fixing GFX Executor missing threadfence_system from v1.67

gilbertlee-amd:
The threadfence_system prior to timestamp collection in v1.67 was accidentally removed which could cause artificially inflated Transfer times because stop timestamp could be collected prior to all writes being drained.  This should not have impacted GFX Executor times.

### DIFF
--- a/src/header/TransferBench.hpp
+++ b/src/header/TransferBench.hpp
@@ -92,7 +92,7 @@ namespace TransferBench
   using std::set;
   using std::vector;
 
-  constexpr char VERSION[] = "1.67";
+  constexpr char VERSION[] = "1.68";
 
   /**
    * Enumeration of supported Executor types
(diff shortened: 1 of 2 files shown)


nileshnegi (approved):

src/header/TransferBench.hpp
```diff
@@ -5459,12 +5465,14 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
...
-            }
+	    if (exeInfo.subExecParamCpu[subExecIdx].N != 0) {
+	      minStartCycle = std::min(minStartCycle, exeInfo.subExecParamGpu[subExecIdx].startCycle);
+	      maxStopCycle  = std::max(maxStopCycle,  exeInfo.subExecParamGpu[subExecIdx].stopCycle);
+	      if (cfg.general.recordPerIteration) {
+		CUs.insert(std::make_pair(exeInfo.subExecParamGpu[subExecIdx].xccId,
+					  GetId(exeInfo.subExecParamGpu[subExecIdx].hwId)));
+	      }
+	    }
           }
```
nileshnegi, line 5477: alignment issue?

PR #253, opened 2026-04-08
Dynamic Loading IB verbs

AtlantaPepsi:
## Motivation
- Dynamically loading ib verbs symbols
- Minor refactor on dmabuf detection and export

## Technical Details
- Created a top level `third-party/` folder for ibverbs related files. Will also harbor future external source which TransferBench depends
- Created a separate minimal header `IbvHeader.hpp` for ib verbs structs and `IbvDynLoad.hpp` for dynamic loading and status report for ib verbs functionality. 
- Dynamic loading is a singleton and done once per process, and TransferBench header will probe in runtime if basic ibverbs function as well as dmabuf export is supported.
- Also got rid `HAVE_DMABUF_SUPPORT` macro. Got rid of redundant dependency check on hsa header and rocr binaries (they are mandatory for AMD platform) in build process. Similar to ibv, it now dynamically checks for `hsa_amd_portable_export_dmabuf` symbol as part of check kernel support, and returns dmabuf support in runtime.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #331, opened 2026-07-08
1.69 feature merge into candidate

AtlantaPepsi:
## Motivation
Given recent fixes that became 1.68 release, we are promoting features originally planned for 1.68 to 1.69.
Rerouting changes on `candidate-1.68`to the `candidate` branch.

## Technical Details
- IB verbs dynamic loading, see #253 
- TDM integration into TransferBench, including @alex-breslow-amd 's initial prototype #322 plus subsequent clean ups, pending further testing and changes.

## Test Plan

ibverbs dynamic loading have been tested fully, including packaging on no-RDMA platform and unpackaging+dynamic loading on IB/RoCE platforms.

TDM's initial results were tested on gfx1250 platforms. Subsequent changes only reports no issues from compilers and linters, testing TODO. 

## Submission Checklist


src/header/TransferBench.hpp
```diff
@@ -5703,6 +5783,160 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
...
+    // which partitions both the work and the staging buffer across every warp in the block
+    // and handles head/tail alignment internally.
+    size_t const sizeBytes = p.N * sizeof(float);
+
+    int subIterations = 0;
+    while (1) {
+      tdm::tdmCopy(dst, src, sizeBytes, shmem, ldsBytes);
+      if (++subIterations == numSubIterations) break;
+    }
+
```
alex-breslow-amd, line 5847: After the while due to the semantics of how waiting is done with the tensor wait API that calls __builtin_amdgcn_tensorcnt(0), it can happen that the first warp completes all of its works before the other warps do and can enter the if(shouldRecordTiming) block.  However, this is invalid because not all other warps have completed their work.

What I would suggest is one of two options:
1) Change the tensor wait API call to support a scope (warp or block).  If block, call __syncthreads after the __builtin_amdgcn_tensorcnt(0) call.  
2) Implement it directly in the kernel.

__builtin_amdgcn_tensorcnt(0);
__syncthreads();

//  Option 1) Release fence at system scope on the sender
__builtin_amdgcn_tensorcnt(0);  // a)
__syncthreads(); // b)
__builtin_amdgcn_fence(__ATOMIC_RELEASE, "");

a) and b) can just be your implementation of block scope sync for TDM

After this, you need one of two things.  Either __threadfence_system or you need to set the scope bits in the actual intrinsic for the TDM loads and stores to be system scope, and then you don't need this.  The block level waiting is enough.

In general TDM with a cache policy of 0, is workgroup processor scope with regular temporal behavior.  This means that the writes made by TDM to the remote GPU's memory are not guaranteed to make it unless you have the explicit release fence mentioned above, which causes the L2 cache writeback, which pushes the data out to the remote GPU.  

However, if the writes made by TDM are marked as system scope by properly setting the cp field, then the writes will write through to the remote memory on the remote GPU.

TransferBench historically has relied on the cache flush that occurs when a kernel ends for correctness.  Recent changes to TransferBench fix this once I brought that up.  You'll notice that the new in kernel timestamping logic now has a __threadfence_system before it in the gfx kernel.

src/header/tdmCopy.h
```diff
@@ -0,0 +1,495 @@
...
+}
+__device__ inline void store(const gfx1250_TDM_GROUP0& g0,
+                             const gfx1250_TDM_GROUP1& g1) {
+    __builtin_amdgcn_tensor_store_from_lds(
+        g0.m_bitfield,
+        g1.m_bitfield,
+        u32x4{}, u32x4{}, u32x8{},
+        /*cpol=*/0);
+}
+__device__ inline void waitTensor0() { __builtin_amdgcn_s_wait_tensorcnt(0); }
```
alex-breslow-amd, line 339: It's fine to keep this warp level for most of the data copies because for performance we actually want to have warps get out of sync and run ahead if they can.  However, at the end just before we collect timing, we need to verify that the writes made by TDM have completed.  See my prior comment.

alex-breslow-amd: Synced internally and decided not to move forward with pipelining so this is enough.

src/header/tdmCopy.h
```diff
@@ -0,0 +1,495 @@
...
+using u32x4 = __attribute__((ext_vector_type(4))) uint32_t;   // D0, D2, D3
+using u32x8 = __attribute__((ext_vector_type(8))) uint32_t;   // D1, D4
+
+__device__ inline void load(const gfx1250_TDM_GROUP0& g0,
+                            const gfx1250_TDM_GROUP1& g1) {
+    __builtin_amdgcn_tensor_load_to_lds(
+        g0.m_bitfield,               // D0  addresses
+        g1.m_bitfield,               // D1  2D shape
+        u32x4{}, u32x4{}, u32x8{},   // D2/D3/D4  higher dims: unused (zero)
+        /*cpol=*/0);
```
alex-breslow-amd, line 329: Set cache policy to system *cpol* field.  Example here on how to compute the value: https://github.com/ROCm/rocm-systems/pull/7019

Follow up with me if you have questions.  Otherwise, you will need a __threadfence_system for in-kernel timing to be accurate.

src/header/TransferBench.hpp
```diff
@@ -5703,6 +5783,160 @@ static bool IsConfiguredGid(union ibv_gid const& gid)
...
+    // Retained for now but unused: tdmCopy fixes the descriptor data-size (4-byte word)
+    // and cache policy internally.
+    (void)dataSize;
+    (void)temporalMode;
+
+    int64_t startCycle;
+    bool const shouldRecordTiming = (threadIdx.x == 0);
+    if (shouldRecordTiming) startCycle = GetTimestamp();
+
+    extern __shared__ __align__(128) float shmem[];
```
alex-breslow-amd, line 5825: Don't need the __align__(128), I added this but it turns out that it's not strictly necessary.  Up to you to keep.

src/header/TransferBench.hpp
gilbertlee-amd: This isn't quite true.  This was just a miss in v1.67.  Versions before that had the threadfence_system.

PR #334, opened 2026-07-14
[debug] option to disable validation

nileshnegi:
## Motivation

Adds a performance debug option that disables validation, so only the initialization and actual transfers execute.

Previously, `ALWAYS_VALIDATE` was a boolean (0 = validate once at the end, 1 = validate every iteration), with no way to skip validation entirely. This makes it difficult to isolate pure transfer/initialization cost from validation overhead.

## Technical Details

Adds a "disable" option to `ALWAYS_VALIDATE` while preserving the existing 0/1 behavior (backward compatible):

- `ALWAYS_VALIDATE=-1` — validation **disabled** (init + transfers only)
- `ALWAYS_VALIDATE=0` — validate **once at the end** (default, unchanged)
- `ALWAYS_VALIDATE=1` — validate **after each iteration** (unchanged)

Changes:
- `src/header/TransferBench.hpp`: per-iteration validation gated on `== 1`; end-of-run validation gated on `== 0`; disabled when `== -1`.
- `src/client/EnvVars.hpp`: updated field comment, help text, and runtime print description for the three modes.
- `src/client/Presets/SmokeTest.hpp`: continues to default to `1` (validate each iteration).
- Interactive mode (`USE_INTERACTIVE=1`): validation is now a separate stage requiring an `<Enter>` keypress rather than running as part of the transfer stage. It respects `ALWAYS_VALIDATE` — skipped for `-1`, `<Enter>`-gated for `0`, and skipped for `1` (which already validates inline each iteration).

## Test Plan

- Built with `make clean && make -j`.
- Ran `./TransferBench cmdline 1K "1 1 g0 d0 g1"` across `ALWAYS_VALIDATE={-1,0,1}` and `USE_INTERACTIVE={0,1}`.

## Test Result

- `ALWAYS_VALIDATE=-1` — no validation output; only init + transfers execute (including under `USE_INTERACTIVE=1`). Confirmed with `AMD_LOG_LEVEL` that there's no CopyDeviceToHost at the end.
- `ALWAYS_VALIDATE=0` (default) — single validation at end; in interactive mode it waits for `<Enter>` before validating and printing PASS/FAIL results.
- `ALWAYS_VALIDATE=1` — validates each iteration inline; no extra Enter-gated stage in interactive mode.
- All transfers reported `PASS`; builds clean with no new warnings.
 
## Submission Checklist

### DIFF
--- a/src/client/EnvVars.hpp
+++ b/src/client/EnvVars.hpp
@@ -588,7 +588,7 @@ public:
     return defaultValue;
   }
 
-  // Validates ALWAYS_VALIDATE is one of the supported modes (-1/0/1); warns and clamps otherwise
+  // Clamps ALWAYS_VALIDATE to a supported mode (-1/0/1), warning on invalid input
   static int NormalizeValidateMode(int value)
   {
     if (value < -1 || value > 1) {
--- a/src/client/EnvVars.hpp
+++ b/src/client/EnvVars.hpp
@@ -79,7 +79,7 @@ public:
   int useInteractive;                // Pause for user-input before starting transfer loop
 
   // Data options
-  int alwaysValidate;                // 0 = disable validation, 1 = validate once after all iterations, 2 = validate after each iteration
+  int alwaysValidate;                // -1 = disable validation, 0 = validate once after all iterations, 1 = validate after each iteration
   int blockBytes;                    // Each subexecutor, except the last, gets a multiple of this many bytes to copy
   int byteOffset;                    // Byte-offset for memory allocations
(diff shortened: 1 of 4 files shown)


gilbertlee-amd (changes requested):
With alwaysValidate=0, we can speed things up even more and skip prepare the dstReference arrays.  

For now, let's have:
 < 0 = no validation
0 = validate at end (default)
> 0 = validate per iteration

gilbertlee-amd (approved):

PR #343, opened 2026-07-21
Add GPU-initiated and Host-initiated SDMA executors via Anvil

nileshnegi:
## Motivation

TransferBench can drive SDMA copies from the host (`DMA`/`BMA` executors), but has no way to benchmark **GPU-initiated** SDMA, where a GPU kernel itself builds SDMA packets and rings the KFD queue doorbell.

This PR adds two new SDMA executors inspired by the rocm-xio "anvil" SDMA queue layer:
- **`GMA`** (`EXE_GPU_INITIATED_DMA`, char code `S`) => a GPU kernel builds the SDMA packet and rings the KFD doorbell (GPU-initiated).
- **`HMA`** (`EXE_HOST_INITIATED_DMA`, char code `H`) => a CPU-driven variant of the same anvil path, where the host builds packets, rings the doorbell, and polls for completion (a correctness/perf reference for the kernel path, and useful in its own right).

Both are measured with the same size sweeps, concurrency, and validation as the existing executors.

The PR also adds the supporting pieces needed to make these paths correct and usable for large transfers (copy chunking, an optional fused copy+signal packet on gfx1250), extends the `AllToAll` and `PodAllToAll` presets to select all executors, and addresses reviewer feedback on BMA batching and the vendored anvil layer.

## Technical Details

* **New `GMA` executor** (GPU-initiated SDMA)
   * Adds `EXE_GPU_INITIATED_DMA` (char `S`) to the executor enum / `ExeTypeStr` (`"CGDINBTSH"`) / `IsGpuExeType` dispatch, `ExeTypeToStr` (`"GMA"`), and HSA-agent resolution.
   * `PrepareAnvilExecutor` / `RunAnvilExecutor` / teardown: one KFD SDMA queue and completion signal per transfer, launched with `hipExtLaunchKernelGGL` for event-bracketed timing; transfers run concurrently on their own streams.
   * The SDMA kernel is split into four non-branching variants (signal/legacy × wait/no-wait); runtime toggles: `ANVIL_WAIT_SIGNAL`, `ANVIL_LEGACY`, `ANVIL_REMOTE_DOORBELL`, opt-in round-robin SDMA engine selection (`ANVIL_ENGINE_ROUND_ROBIN`), and an XGMI-optimal SDMA affinity topology table.
* **New `HMA` executor** (host-initiated SDMA)
   * Adds `EXE_HOST_INITIATED_DMA` (char `H`), driven by `RunAnvilHostExecutor` => promoted from the earlier `ANVIL_HOST_QUEUE=1` env toggle into a first-class executor.
   * `SdmaQueueHostHandle`: the host reserves/wrap-pads/places packets into the same host-accessible KFD ring, advances the wptr, rings the doorbell, and polls `quiet()`. Channels are keyed by `{src,dst}`; `sdma_packets.hpp` provides the host-side packet builders. A queue is driven by either the host handle or the device kernel, never both at once.
* **anvil library** (`src/anvil/`)
   * inspired by rocm-xio `sdma-ep`; gated behind `ANVIL_EXEC_ENABLED` (CMake `-DENABLE_ANVIL_EXEC`, Makefile `ENABLE_ANVIL_EXEC`), linking `libhsakmt` + libdrm transitively.
* **Chunking (>1 GiB transfers)**
   * the SDMA `COPY_LINEAR` count field is 30-bit (stores size-1), capping one packet at 1 GiB.
   * `put_chunked` / `put_signal_chunked` split larger transfers into ≤1 GiB packets, with the signal on the final chunk only.
   * size is controlled by `ANVIL_CHUNK_SIZE` (raw bytes or K/M/G, clamped to `(0, 1 GiB]`).
* **OSS7 fused copy+signal (gfx1250)**
   * an optional single fused `COPY_LINEAR_WAIT_SIGNAL_MI4` packet replacing the separate `COPY_LINEAR` + `ATOMIC` pair, selected via `ANVIL_FUSED_SIGNAL=1`.
   * Two-tier gating: `XIO_SDMA_OSS7` (build-level, ABI-portable structs/host code) and `XIO_SDMA_OSS7_ENABLED` (arch-gates the fused device code to gfx1250/gfx950), so all-arch package builds are safe; the fused path is selected at runtime only on gfx1250.
   * On by default; `-DENABLE_XIO_SDMA_OSS7=OFF` forces the separate path.
* **`AllToAll` / `PodAllToAll` executor selection**
   * `USE_DMA_EXEC` now selects the executor in both presets: `0=GFX, 1=DMA, 2=BMA, 3=GMA, 4=HMA`.
   * GMA/HMA (`3`/`4`) require `-DENABLE_ANVIL_EXEC=ON` and are rejected on multi-rank runs (anvil uses process-local virtual addresses, so it cannot reach another rank's memory).
* **BMA batching** (addresses review feedback)
   * BMA (`USE_DMA_EXEC=2`) now batches every copy from a given source GPU into a **single** `hipMemcpyBatchAsync` launch by emitting one multi-destination `Transfer` per source GPU, rather than launching each `(src,dst)` copy independently.
   * As noted in review, per-link timing can no longer be measured individually; each `(src,dst)` cell instead shows its equal share of the measured per-source aggregate (rows sum to the source total, diagonal stays `N/A`). `USE_REMOTE_READ` is ignored for BMA (the source GPU is always the executor).
* **Correct byte accounting for multi-destination transfers**
   * An executor's byte total (and hence CPU-timed bandwidth) now scales by the number of destinations, so a merged BMA batch that writes `numBytes` to each of K destinations is counted correctly instead of being under-counted by ~K×.
* **Vendored anvil review fixes**
   * Fix `CHECK_HSAKMT_SUCCESS` double-evaluation of its `call` argument (capture the result once).
   * Cast `std::tolower` arguments to `unsigned char` (avoid UB on negative `char`), add `<cctype>`.
   * Fix `AnvilLib& anvil = anvil.getInstance();` self-reference → `AnvilLib::getInstance()`.
   * Add `<iosfwd>` for the `dump(std::ofstream&)` declaration.
* **Diagnostics/tuning**
   * `ANVIL_CPU_TIMING=1` reports CPU wall time next to the hipEvent time to quantify launch+sync overhead.

## Test Plan

* Build with the anvil executor enabled for gfx1250 (`-DENABLE_ANVIL_EXEC=ON -DGPU_TARGETS=gfx1250`), default `XIO_SDMA_OSS7=ON`.
* Run GMA G→G transfers across a size sweep, including sizes >1 GiB, to exercise the chunking path.
* Force small `ANVIL_CHUNK_SIZE` values to validate multi-chunk copies produce a single completion signal and correct data.
* Validate correctness with `FILL_PATTERN` + always-validate.
* Exercise the HMA (host-initiated) executor (`USE_DMA_EXEC=4`) and the host-queue API unit harness (`tools/anvil_harness.cpp`).
* Exercise the fused path (`ANVIL_FUSED_SIGNAL=1`) on gfx1250 and confirm the non-gfx1250 runtime fallback.
* Run `a2a` and `poda2a` across all executors (`USE_DMA_EXEC=0..4`) and confirm BMA merged batching produces correct per-source aggregates and CPU-timed bandwidth in line with GPU-timed.

## Test Result

* Clean build for gfx1250 with `ENABLE_ANVIL_EXEC=ON` and `XIO_SDMA_OSS7=ON`.
* GiB transfers complete correctly via chunking (single completion signal preserved); small chunk sizes validated for correctness.
* Fused copy+signal validated on gfx1250 (64 KiB - 1 GiB), perf-neutral vs the separate path; falls back with a log line on non-gfx1250 runtimes.
* `a2a` (4×GPU, 128 MiB) verified across all executors with no GPU faults; representative aggregate bandwidth (GPU-timed / CPU-timed)

## Submission Checklist

* [ ] Look over the contributing guidelines at
  <https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests>.

### DIFF
--- a/src/client/Presets/AllToAll.hpp
+++ b/src/client/Presets/AllToAll.hpp
@@ -54,7 +54,7 @@ int AllToAllPreset(EnvVars&          ev,
   int numResults    = EnvVars::GetEnvVar("NUM_RESULTS"    , numRanks > 1 ? 1 : 0);
   int numSubExecs   = EnvVars::GetEnvVar("NUM_SUB_EXEC"   , 8);
   int showDetails   = EnvVars::GetEnvVar("SHOW_DETAILS"   , 0);
-  int useDmaExec    = EnvVars::GetEnvVar("USE_DMA_EXEC"   , 0);
+  int useDmaExec    = EnvVars::GetEnvVar("USE_DMA_EXEC"   , 0);  // 0=GFX, 1=DMA, 2=BMA, 3=GISDMA(anvil)
   int useTdmExec    = EnvVars::GetEnvVar("USE_TDM_EXEC"   , 0);
   int useRemoteRead = EnvVars::GetEnvVar("USE_REMOTE_READ", 0);
 
@@ -65,8 +65,14 @@ int AllToAllPreset(EnvVars&          ev,
   }
 
   // Determine which GPU executor to use
-  ExeType     exeType    = useDmaExec ? EXE_GPU_DMA : (useTdmExec ? EXE_GPU_TDM : EXE_GPU_GFX);
-  char const* execName   = useDmaExec ? "DMA" : (useTdmExec ? "TDM" : "GFX");
+  ExeType exeType = (useDmaExec == 3) ? EXE_GPU_INITIATED_DMA :
+                    (useDmaExec == 2) ? EXE_GPU_BDMA :
+                    (useDmaExec == 1) ? EXE_GPU_DMA :
+                    (useTdmExec)      ? EXE_GPU_TDM : EXE_GPU_GFX;
+  char const* execName = (useDmaExec == 3) ? "GISDMA" :
+                         (useDmaExec == 2) ? "BMA" :
+                         (useDmaExec == 1) ? "DMA" :
+                         (useTdmExec)      ? "TDM" : "GFX";
 
   // Check that all ranks have at least the number of GPUs requested
   // Warn if NIC configuration is slightly different from one another
@@ -118,7 +124,10 @@ int AllToAllPreset(EnvVars&          ev,
         ev.Print("NUM_RESULTS"  , numResults   , "Showing top/bottom %d results", numResults);
       ev.Print("NUM_SUB_EXEC"   , numSubExecs  , "Using %d subexecutors/CUs per Transfer", numSubExecs);
       ev.Print("SHOW_DETAILS"   , showDetails  , "%s full Test details", showDetails ? "Showing" : "Hiding");
-      ev.Print("USE_DMA_EXEC"   , useDmaExec   , "Using %s executor", useDmaExec ? "DMA" : "GFX");
+      ev.Print("USE_DMA_EXEC"   , useDmaExec   , "Using %s executor",
+               useDmaExec == 3 ? "GISDMA (GPU-initiated SDMA/anvil)" :
+               useDmaExec == 2 ? "BMA" :
+               useDmaExec == 1 ? "DMA" : "GFX");
       ev.Print("USE_TDM_EXEC"   , useTdmExec   , "Using %s executor", useTdmExec ? "TDM" : "GFX");
       ev.Print("USE_REMOTE_READ", useRemoteRead, "Using %s as executor", useRemoteRead ? "DST" : "SRC");
       printf("\n");
@@ -133,6 +142,16 @@ int AllToAllPreset(EnvVars&          ev,
     Utils::Print("[ERROR] %s execution can only be used for copies (A2A_MODE=0)\n", execName);
     return ERR_FATAL;
   }
+  if (useDmaExec < 0 || useDmaExec > 3) {
+    Utils::Print("[ERROR] USE_DMA_EXEC must be 0 (GFX), 1 (DMA), 2 (BMA), or 3 (GISDMA)\n");
+    return ERR_FATAL;
+  }
+#ifndef ANVIL_EXEC_ENABLED
+  if (useDmaExec == 3) {
+    Utils::Print("[ERROR] USE_DMA_EXEC=3 (GISDMA) requires building with -DENABLE_ANVIL_EXEC=ON\n");
+    return ERR_FATAL;
+  }
+#endif
   if (numResults * 2 > numRanks) {
     Utils::Print("[ERROR] Number of extrema results requested exceeds number of ranks.  NUM_RESULTS should be at most half the number of ranks\n");
     return ERR_FATAL;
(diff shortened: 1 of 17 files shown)


src/client/Presets/AllToAll.hpp
```diff
@@ -117,17 +120,29 @@ int AllToAllPreset(EnvVars&          ev,
...
+#endif
   if (numResults * 2 > numRanks) {
     Utils::Print("[ERROR] Number of extrema results requested exceeds number of ranks.  NUM_RESULTS should be at most half the number of ranks\n");
     return ERR_FATAL;
   }
 
   // Collect the number of GPU devices to use
-  ExeType exeType = useDmaExec ? EXE_GPU_DMA : EXE_GPU_GFX;
+  ExeType exeType = (useDmaExec == 3) ? EXE_GPU_INITIATED_DMA :
+                    (useDmaExec == 2) ? EXE_GPU_BDMA :
```
gilbertlee-amd: BDMA is a good add, however the correct way for this to be handled will be to actually "batch" together multiple "B" Transfers into a single launch, (which unfortunately will make us lose per Transfer information).  We will need to expose this as a new config option for BMA.  This is ok as a place holder for now

src/client/Client.cpp
```diff
@@ -257,7 +257,7 @@ void DisplayVersion()
 #ifndef TB_GIT_COMMIT
 #define TB_GIT_COMMIT "unknown"
 #endif
-  Print("TransferBench v%s.%s (%s:%s)%s%s\n", VERSION, CLIENT_VERSION, TB_GIT_BRANCH, TB_GIT_COMMIT, support.c_str(), multiNodeMode.c_str());
+  Print("TransferBench v%s.%sC (%s:%s)%s%s\n", VERSION, CLIENT_VERSION, TB_GIT_BRANCH, TB_GIT_COMMIT, support.c_str(), multiNodeMode.c_str());
```
gilbertlee-amd: needs fix

src/header/TransferBench.hpp
```diff
@@ -2664,6 +2686,52 @@ namespace {
...
+              "Transfer %d: EXE_GPU_INITIATED_DMA executor GPU (%d) must match source GPU (%d)",
+              i, t.exeDevice.exeIndex, t.srcs[0].memIndex});
+          hasFatalError = true;
+          break;
+        }
+        if (t.srcs[0].memRank != t.exeDevice.exeRank || t.dsts[0].memRank != t.exeDevice.exeRank) {
+          errors.push_back({ERR_FATAL,
+              "Transfer %d: EXE_GPU_INITIATED_DMA does not support cross-rank transfers",
+              i});
+          hasFatalError = true;
```
gilbertlee-amd, line 2718: Do we expect this to work within a pod?

gilbertlee-amd (approved):

src/header/TransferBench.hpp
nileshnegi: added support

src/client/Presets/AllToAll.hpp
nileshnegi: made some mods in newer commits. pls check

PR #350, opened 2026-08-20
Removal of 2D copy

AtlantaPepsi:
## Motivation

Simplification of TDM copy issue API

## Technical Details

Per our experiments we did not find noticeable performance difference for 1D vs 2D copy, until further supporting evidence, we are removing 2D pattern of copy

### DIFF
--- a/src/header/tdmCopy.h
+++ b/src/header/tdmCopy.h
@@ -320,8 +320,8 @@ __device__ TDM_API void tdmWait() TDM_DELETED;
 //   * This copy is single-buffered (one LDS window per warp), so each tile is a
 //     load->store dependency chain on that window: the store reads what the load
 //     wrote (RAW), and the next tile's load overwrites the window the store is
-//     still draining (WAR). Both edges need an s_wait_tensorcnt, so issueRows()/
-//     issueRow1d() wait after the load and after the store. Consequence: with a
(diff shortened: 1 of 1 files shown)


alex-breslow-amd (approved):
LGTM

src/header/tdmCopy.h
```diff
@@ -398,24 +398,21 @@ __device__ inline void warpVecCopy(const uint8_t* s, uint8_t* d, size_t n,
...
+__device__ inline void issueRow(uint64_t src, uint32_t lds, uint64_t dst,
+                                uint32_t rows) {
+    uint32_t dwords = rows * TD0;            // whole chunk as one flat dword run
     gfx1250_TDM_GROUP1 g1;
     g1.dataSize(DS4);
-    g1.tileDim0(TD0);   g1.tileDim1(rows);
-    g1.tensorDim0(TD0); g1.tensorDim1(rows);
-    g1.tensorDim0Stride(TD0);                // rows back-to-back (contiguous)
+    g1.tileDim0(dwords);   g1.tileDim1(1);
+    g1.tensorDim0(dwords); g1.tensorDim1(1);
```
alex-breslow-amd, line 413: I'm curious here if the default setting of tileDim1 to 0 gives the same performance.  Same with tensorDim1.  You may not even have to set these to 1.

PR #352, opened 2026-08-25
Fix NIC to GPU proximity detection on systems with multiple PCIe domains

paklui:
## Motivation

On systems where each GPU sits in its own PCIe domain, TransferBench picked the  wrong NIC as closest, thus every GPU mapped to a single front-end NIC that is not wired to any GPU, while the back-end NICs actually adjacent to each GPU mapped to nothing. The mapping decides which NIC a GPU uses for RDMA transfers (`NIC_NEAREST` executor, `nic_rings` preset). With incorrect mapping it would mean benchmarking ising a NIC that is not attached to that GPU's root complex, producing misleading numbers. The causes are:

1. The proximity metric compared raw bus numbers. Bus numbers repeat across domains, so devices in different domains looked "close" purely because their bus numbers were numerically near.
2. When several NICs were equally close, only one was recorded, chosen by a round-robin counter, so a GPU with two adjacent NICs only ever reported one.

This PR fixes the topology detection so it works correctly on systems with multiple PCIe domains and multiple NICs per GPU (verified on a 4-GPU, 8 BE + 1 FE, or 2 BE NICs per GPU).

## Technical Details

1. Use PCIe domain, not bus number, for the distance metric
Two functions were renamed to match what they now do. `ExtractBusNumber` became `ExtractDomain`, and `GetBusIdDistance` became `GetDomainDistance`.

The distance between two devices is now computed as follows:
- If both devices sit in the same PCIe domain, the distance is 0. Within a single domain, the PCIe tree search already tells us how the two devices are related, so no extra tie-break is needed.
- If the two devices sit in different domains, the distance is the difference between their domain numbers. That value is always at least 1, so a device in a different domain always ranks behind every device in the same domain.
Bus numbers are no longer used to compare devices within a domain. Firmware assigns them, and they do not reliably indicate which device is physically closer. Using them would also separate NICs that are genuinely the same distance from a GPU, and reporting those ties is the point of this fix.

2. Record all equally-close NICs/GPUs
Replaces the `assignedCount` round-robin, which picked one NIC per GPU even when several were tied. Both directions now keep the full tie set, matching how `GetClosestNicsToGpu()` was already shaped.

Trade-off: the removed round-robin was intended to spread multi-port NICs (ports sharing one busId) across GPUs. With ties recorded, `nic_rings` now uses *all* tied ports, better, but a bare `NIC_NEAREST` resolves slot 0 for every GPU rather than spreading. Use the explicit `N<gpu>.<slot>` form to select a specific port.

3. Update the CLI topology display
The display called the scalar `GetClosestNicToGpu()`, so it would have shown one NIC even after (2). Switched to the vector API. The NIC list is space-separated so the field stays a single column under `OUTPUT_TO_CSV=1`, where the separator is a comma.

- `ionic_8`'s port is down, so GPU 3 reports only NIC 7. NICs that cannot execute a Transfer are excluded from the mapping: `NicIsActive()` rejects such a Transfer with `ERR_FATAL` rather than falling back, so one would make `nic_rings` abort on any tray with a down link.
- `OUTPUT_TO_CSV=1` rows hold a constant 11 fields regardless of NIC count.

## Test Plan

Compare both unmodified `develop` (44dec25) and this branch merged with `develop` on the same hardware, and compared the topology output.

- System with 4 GPUs, 8 BE NICs + 1 FE NIC. For example, 2 NICs are wired via PCIe per GPU, each GPU in its own PCIe domain.
- latest therock or ROCm 7.15.0

## Test Result

Current `develop` HEAD:
```
 NIC | Device Name | Active | PCIe Bus ID  | NUMA | Closest GPU(s) | GID Index | GID Descriptor
-----+-------------+--------+--------------+------+----------------+-----------+-------------------
 0   | ionic_0     | Yes    | 0000:04:00.0 | 0    | 0,1,2,3        | 1         | RoCEv2 IPv4-mapped IPv6
 1   | ionic_1     | Yes    | 0001:44:00.0 | -1   |                | 1         | RoCEv2 IPv4-mapped IPv6
 2   | ionic_2     | Yes    | 0001:48:00.0 | -1   |                | 1         | RoCEv2 IPv4-mapped IPv6
 3   | ionic_3     | Yes    | 0003:44:00.0 | -1   |                | 1         | RoCEv2 IPv4-mapped IPv6
 4   | ionic_4     | Yes    | 0003:48:00.0 | -1   |                | 1         | RoCEv2 IPv4-mapped IPv6
 5   | ionic_5     | Yes    | 0002:44:00.0 | -1   |                | 1         | RoCEv2 IPv4-mapped IPv6
 6   | ionic_6     | Yes    | 0002:48:00.0 | -1   |                | 1         | RoCEv2 IPv4-mapped IPv6
 7   | ionic_7     | Yes    | 0004:44:00.0 | -1   |                | 1         | RoCEv2 IPv4-mapped IPv6
 8   | ionic_8     | No     | 0004:48:00.0 | -1   |                | N/A       | N/A

        | gfx1250 | gfx1250 | gfx1250 | gfx1250 |
        | GPU 00 | GPU 01 | GPU 02 | GPU 03 | PCIe Bus ID  | #CUs | NUMA | #DMA | #XCC | NIC
--------+--------+--------+--------+--------+--------------+------+------+------+------+------
 GPU 00 |    N/A | XGMI-1 | XGMI-1 | XGMI-1 | 0001:01:00.0 | 256  | 0    | 16   | 8    | 0
 GPU 01 | XGMI-1 |    N/A | XGMI-1 | XGMI-1 | 0002:01:00.0 | 256  | 0    | 16   | 8    | 0
 GPU 02 | XGMI-1 | XGMI-1 |    N/A | XGMI-1 | 0003:01:00.0 | 256  | 0    | 16   | 8    | 0
 GPU 03 | XGMI-1 | XGMI-1 | XGMI-1 |    N/A | 0004:01:00.0 | 256  | 0    | 16   | 8    | 0
```

Currently every one of the 4 GPUs is mapped to NIC 0 which is on PCIe domain 0000, a completely different domain from any GPU (`0001`-`0004`). NIC 0 is a Front End NIC that is not physically wired to any GPU. Meanwhile the 8 Back End NICs that are actually adjacent to a GPU (`ionic_1`-`ionic_8`) show up as closest to blank. This is the domain-blind bus-number comparison from change (2) picking the numerically-smallest bus number regardless of domain, combined with (3)'s single-NIC-only recording.

actual output from ./TransferBench after (this branch merged with `develop`):

```
 NIC | Device Name | Active | PCIe Bus ID  | NUMA | Closest GPU(s) | GID Index | GID Descriptor
-----+-------------+--------+--------------+------+----------------+-----------+-------------------
 0   | ionic_0     | Yes    | 0000:04:00.0 | 0    |                | 1         | RoCEv2 IPv4-mapped IPv6   ### NOTE: FE NIC, not GPU-facing
 1   | ionic_1     | Yes    | 0001:44:00.0 | -1   | 0              | 1         | RoCEv2 IPv4-mapped IPv6
 2   | ionic_2     | Yes    | 0001:48:00.0 | -1   | 0              | 1         | RoCEv2 IPv4-mapped IPv6
 3   | ionic_3     | Yes    | 0003:44:00.0 | -1   | 2              | 1         | RoCEv2 IPv4-mapped IPv6
 4   | ionic_4     | Yes    | 0003:48:00.0 | -1   | 2              | 1         | RoCEv2 IPv4-mapped IPv6
 5   | ionic_5     | Yes    | 0002:44:00.0 | -1   | 1              | 1         | RoCEv2 IPv4-mapped IPv6
 6   | ionic_6     | Yes    | 0002:48:00.0 | -1   | 1              | 1         | RoCEv2 IPv4-mapped IPv6
 7   | ionic_7     | Yes    | 0004:44:00.0 | -1   | 3              | 1         | RoCEv2 IPv4-mapped IPv6
 8   | ionic_8     | No     | 0004:48:00.0 | -1   | 3              | N/A       | N/A            

        | gfx1250 | gfx1250 | gfx1250 | gfx1250 |
        | GPU 00 | GPU 01 | GPU 02 | GPU 03 | PCIe Bus ID  | #CUs | NUMA | #DMA | #XCC | NIC
--------+--------+--------+--------+--------+--------------+------+------+------+------+------
 GPU 00 |    N/A | XGMI-1 | XGMI-1 | XGMI-1 | 0001:01:00.0 | 256  | 0    | 16   | 8    | 1,2
 GPU 01 | XGMI-1 |    N/A | XGMI-1 | XGMI-1 | 0002:01:00.0 | 256  | 0    | 16   | 8    | 5,6
 GPU 02 | XGMI-1 | XGMI-1 |    N/A | XGMI-1 | 0003:01:00.0 | 256  | 0    | 16   | 8    | 3,4
 GPU 03 | XGMI-1 | XGMI-1 | XGMI-1 |    N/A | 0004:01:00.0 | 256  | 0    | 16   | 8    | 7,8
```

Each GPU is now correctly mapped to the NIC(s) in its own PCIe domain: 
- GPU 0 -> NICs 1,2 (domain `0001`)
- GPU 1->NIC 5,6 (domain `0002`)
- GPU 2 -> NICs 3,4 (domain `0003`)
- GPU 3 -> NICs 7,8 (domain `0004`)
- NIC 0 (domain 0000, FE NIC) correctly shows no GPU mapping, since no GPU exists in its domain

Also confirmed a clean build with no new warnings, using the dynamically-loaded libibverbs path (`IsIbvSymbolsReady()`) that `develop` added since this branch's base, e.g. this fix is verified against the current `develop` HEAD

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


gilbertlee-amd:
Merged as part of 1.70

PR #357, opened 2026-09-14
TransferBench v1.70.02

AtlantaPepsi:
## Motivation

- v1.70.01 Skipped due to incorrect minor tagging on v1.70.00
- Update minor version tags.
- Modify `rings` to use all available executors by default
- Disable `smoketest` tests that don't scale well in multi-node, i.e. all-to-all, gather, broadcast

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -30,11 +30,8 @@ jobs:
       - name: Install OS requirements
         timeout-minutes: 10
         run: |
-          apt update
-          apt install -y software-properties-common
-          apt-add-repository -y ppa:git-core/ppa
           apt-get update
-          apt install -y git
+          DEBIAN_FRONTEND=noninteractive apt-get install -y git
 
       - name: Checkout repository
         uses: actions/checkout@v4


gilbertlee-amd (approved):

src/client/Presets/SmokeTest.hpp
```diff
@@ -269,9 +269,17 @@ int SmokeTestPreset(EnvVars&          ev,
   MemType cpuMemType = Utils::GetCpuMemType(cpuMemTypeIdx);
   MemType gpuMemType = Utils::GetGpuMemType(gpuMemTypeIdx);
   std::set<int> testsToRun(testList.begin(), testList.end());
+  bool skipCollectiveByDefault = false;
   if (testList.empty()) {
     for (int testIdx = 1; testIdx <= NUM_SMOKE_TESTS; testIdx++)
       testsToRun.insert(testIdx);
+    // Multi-node default: H2D/D2H/D2D only.
+    // Broadcast, gather, and all-to-all remain available via TEST_LIST
```
nileshnegi, line 277: imo, we can keep 13, 14, 15, 16 for multinode

AtlantaPepsi: @gilbertlee-amd thoughts? we originally wanted to disable GFX as well as CU count is finite and might not scale well.

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 Documentation for TransferBench is available at
 [https://rocm.docs.amd.com/projects/TransferBench](https://rocm.docs.amd.com/projects/TransferBench).
 
+## v1.70.02
+### Modified
+- rings preset defaults `NUM_SUB_EXEC` to 0, which uses all available subexecutors per Transfer
+- smoketest disables broadcast, gather, and all-to-all by default on multi-node. Set `TEST_LIST` to enable them
+
+## v1.70.01
+SKIPPED
```
gilbertlee-amd: "Skipped due to misnaming of 1.70.00"

src/client/Presets/SmokeTest.hpp
gilbertlee-amd: Yes.  CU count is finite, and we are switch-based, so I don't think we need to keep 13,14,15,16 for now.  If people want to run full tests, they still can if they specify it explicitly.

gilbertlee-amd (changes requested):

nileshnegi (approved):

gilbertlee-amd (approved):

